PR #1, opened 2015-06-26
New build structure

abellotti:
- Update bin scripts to be /build/manageiq/bin
- Update documentation to reflect cloning manageiq-appliance-build
  and setup a symlink from /build/bin to /build/manageiq/bin

### DIFF
--- a/bin/centos71-devbuild.sh
+++ b/bin/centos71-devbuild.sh
@@ -44,9 +44,9 @@ if [ "${1}" = "--fg" ]
 then
   echo "CentOS 7.1 Dev Build kicked off, Log being saved in ${LOG_FILE} ..."
 
-  time ruby /build/manageiq/scripts/vmbuild.rb $BUILD_OPTIONS 2>&1 | tee ${LOG_FILE}
+  time ruby /build/scripts/vmbuild.rb $BUILD_OPTIONS 2>&1 | tee ${LOG_FILE}
 else
-  nohup time ruby /build/manageiq/scripts/vmbuild.rb $BUILD_OPTIONS >${LOG_FILE} 2>&1 &
+  nohup time ruby /build/scripts/vmbuild.rb $BUILD_OPTIONS >${LOG_FILE} 2>&1 &
 
   echo "CentOS 7.1 Dev Build kicked off, Log @ ${LOG_FILE} ..."
 fi
--- a/bin/centos71-devbuild.sh
+++ b/bin/centos71-devbuild.sh
@@ -44,9 +44,9 @@ if [ "${1}" = "--fg" ]
 then
   echo "CentOS 7.1 Dev Build kicked off, Log being saved in ${LOG_FILE} ..."
 
-  time ruby /build/scripts/vmbuild.rb $BUILD_OPTIONS 2>&1 | tee ${LOG_FILE}
+  time ruby /build/manageiq/scripts/vmbuild.rb $BUILD_OPTIONS 2>&1 | tee ${LOG_FILE}
 else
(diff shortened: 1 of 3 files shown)


docs/creating_build_vm.txt
```diff
@@ -71,9 +69,14 @@ Create /build
...
   /references
-  /scripts
   /storage
 
+Clone the build scripts
+
+  cd /build
+  git clone git@github.com:manageiq/manageiq-appliance-build manageiq
+  ln -s manageiq/bin bin
+
```
jrafanie: @abellotti 

Also:

`ln -s manageiq/scripts scripts` 

right?

kbrock:
Yay. introducing structure

bin/centos71-devbuild.sh
```diff
@@ -44,9 +44,9 @@ if [ "${1}" = "--fg" ]
 then
   echo "CentOS 7.1 Dev Build kicked off, Log being saved in ${LOG_FILE} ..."
 
-  time ruby /build/scripts/vmbuild.rb $BUILD_OPTIONS 2>&1 | tee ${LOG_FILE}
+  time ruby /build/manageiq/scripts/vmbuild.rb $BUILD_OPTIONS 2>&1 | tee ${LOG_FILE}
```
chessbyte: Should this be `manageiq-appliance` where the scripts are moving to?

docs/creating_build_vm.txt
abellotti: wouldn't hurt for convenience, I just pushed up the change.  Thanks.

bin/centos71-devbuild.sh
abellotti: Does not matter namewise, if we require the symlink from above, We can also go back to the way it was.

docs/creating_build_vm.txt
```diff
@@ -71,9 +69,15 @@ Create /build
...
   /references
-  /scripts
   /storage
 
+Clone the build scripts
+
+  cd /build
+  git clone git@github.com:manageiq/manageiq-appliance-build manageiq
+  ln -s manageiq/bin bin
+  ln -s manageiq/scripts bin
```
jrafanie: typo: 
`ln -s manageiq/scripts scripts`

bin/centos71-devbuild.sh
jrafanie: > Should this be manageiq-appliance where the scripts are moving to?

hehe, confusion already... these build scripts live in `manageiq-appliance-build` now.  

It doesn't matter what we call the directory here but it would make more sense for a new person to understand it.  git clone would do the right thing since it will clone it to /build/manageiq-appliance-build

Fryguy: I think removing the symlinks in the end is the best idea.  The symlinks are good in the interim when you are getting this working, so you can flip back and forth, but in the end they are just confusing noise.

docs/creating_build_vm.txt
```diff
@@ -71,9 +69,16 @@ Create /build
   /kickstarts
   /logs
   /references
-  /scripts
   /storage
 
+Clone the build scripts
+
+  cd /build
+  git clone git@github.com:manageiq/manageiq-appliance-build.git
```
jrafanie: oh man, sorry...

we probably want to do this:
```
git clone -o upstream https://www.github.com/ManageIQ/manageiq-appliance-build.git
```

This way we avoid needing ssh keys on imagefactory vm, I think.
And, if you have a fork of the manageiq-appliance-build repo, you can add your origin.

jrafanie:
Looks great for now.  I'm sure we'll want to change this as we play with it more.

PR #2, opened 2015-06-26
Test build/system directory to repo extraction

jrafanie:
Summary of changes:

* Before, the build code worked with a single manageiq repository containing the build and system directories.
* Now, it works with 3 separate repositories (manageiq, manageiq-appliance (former system), and manageiq-appliance-build (former build)
* CLI now accepts 3 repos via -A(--appliance_url), -B(--build_url), and -M(--manageiq_url) or defaults to the ones in the manageiq organization.  The repo is not read from the targets.yml anymore.  The `--repo` option in the CLI was removed as it's now unclear with 3 repos.
* Note:  all 3 repos must have the `--reference` so you need to create branches and tags on each repo to do a build.
* Destination appliance will have manage-appliance checked out in /var/www/manageiq-appliance with a symlink from /var/www/miq/system -> /var/www/manageiq-appliance (for backward compatility)

Example test build command line:

```
ruby /build/scripts/vmbuild.rb -A https://github.com/jrafanie/manageiq-appliance.git -B https://github.com/jrafanie/manageiq-appliance-build.git -M https://github.com/jrafanie/manageiq.git --reference test_build_system_repo_extraction
```

@abellotti @chessbyte Please review.

Note, the appliances build and I get the error during rake assets:precompile because of this: https://github.com/ManageIQ/manageiq/issues/3285, @kbrock is trying to solve that one.

Workers do start properly after reboot though.

Once this is merged, we can then review/merge: https://github.com/ManageIQ/manageiq/issues/3295

### DIFF
--- a/scripts/kickstart_generator.rb
+++ b/scripts/kickstart_generator.rb
@@ -10,14 +10,15 @@ module Build
     KS_DIR     = "kickstarts"
     KS_GEN_DIR = "#{KS_DIR}/generated"
 
-    attr_reader :targets, :puddle, :git_checkout
+    attr_reader :targets, :puddle, :appliance_checkout, :code_checkout
 
-    def initialize(build_base, targets, puddle, git_checkout)
-      @build_base   = Pathname.new(build_base)
-      @ks_gen_base  = @build_base.join(KS_GEN_DIR)
-      @targets      = targets
-      @puddle       = puddle # used during ERB evaluation
-      @git_checkout = git_checkout
+    def initialize(build_base, targets, puddle, code_checkout, appliance_checkout)
+      @build_base         = Pathname.new(build_base)
+      @ks_gen_base        = @build_base.join(KS_GEN_DIR)
+      @targets            = targets
+      @puddle             = puddle # used during ERB evaluation
+      @appliance_checkout = appliance_checkout
+      @code_checkout      = code_checkout
     end
 
     def run(task = :all)
--- a/scripts/kickstart_generator.rb
+++ b/scripts/kickstart_generator.rb
@@ -10,15 +10,15 @@ module Build
     KS_DIR     = "kickstarts"
     KS_GEN_DIR = "#{KS_DIR}/generated"
 
-    attr_reader :targets, :puddle, :appliance_checkout, :code_checkout
+    attr_reader :targets, :puddle, :appliance_checkout, :manageiq_checkout
 
-    def initialize(build_base, targets, puddle, code_checkout, appliance_checkout)
+    def initialize(build_base, targets, puddle, manageiq_checkout, appliance_checkout)
       @build_base         = Pathname.new(build_base)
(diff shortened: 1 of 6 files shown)


Fryguy:
Can one pass appliance_url as a path?  I think it will work if you are git cloning it...it allows test builds from local code, which would be cool.

jrafanie:
@Fryguy it would be possible with build_url since the build vm checks out that code for the kickstart/targets.yml but not for manageiq_url and appliance_url since those are cloned within the nested vm.

Fryguy:
Ah ok, that's interesting and makes sense

scripts/kickstart_generator.rb
```diff
@@ -10,14 +10,15 @@ class KickstartGenerator
...
-    attr_reader :targets, :puddle, :git_checkout
+    attr_reader :targets, :puddle, :appliance_checkout, :manageiq_checkout
 
-    def initialize(build_base, targets, puddle, git_checkout)
-      @build_base   = Pathname.new(build_base)
-      @ks_gen_base  = @build_base.join(KS_GEN_DIR)
-      @targets      = targets
-      @puddle       = puddle # used during ERB evaluation
-      @git_checkout = git_checkout
+    def initialize(build_base, targets, puddle, manageiq_checkout, appliance_checkout)
```
Fryguy: I don't know why, but the ordering bothers me that the attr_reader above and lines 20-21 below have appliance before manageiq, but these params are the other way around.

jrafanie:
@Fryguy Note, the :confused: part:  build_url is weird because it affects the kickstart/targets.yml but not vmbuild.rb itself (since it's the one doing the cloning)

Fryguy:
This is really great stuff, @jrafanie !

jrafanie:
@abellotti Note, I really don't like having to put the whole URL on CLI, makes we wonder if we should eventually allow for slugs like

```
ruby /build/scripts/vmbuild.rb -A "jrafanie/manageiq-appliance" -B "jrafanie/manageiq-appliance-build" -M "jrafanie/manageiq" --reference test_build_system_repo_extraction
```

jrafanie:
Don't merge yet, I will fixup the bin shell scripts to call vmbuild.rb using the new CLI options.

abellotti:
@jrafanie  that would simplify things, it was different when there was only one url on the cli.  Note, optionally just the fork name since the repo names would be the same, i.e. just -A "jrafanie" -B "jrafanie" -M "jrafanie".

scripts/kickstart_generator.rb
jrafanie: ok, i'll fix it.

jrafanie:
Yeah @abellotti agreed, fork name would be better too.  I'm sure we'll find good ideas by trying to fix the callers of vmbuild.rb to use the new -A, -B, -M options, such as: bin/centos71-devbuild.sh

@Fryguy fixed the param list to be a consistent order.

PR #3, opened 2015-06-26
Removed older scripts from the root of the new repo

abellotti:
These scripts are no longer used.  We used the ones in the scripts/ sub-directory.


jrafanie:
:tada: :sparkles: :scissors:

1_8_7_hacks.rb
```diff
@@ -1,20 +0,0 @@
-# The current imagefactory instance has ruby 1.8.7
-# Once we're off of building on that machine, DELETE ME!
```
jrafanie: Yay!  I was looking forward to this day when I wrote that comment.

Fryguy: :+1: :laughing:

PR #4, opened 2015-07-06
Add symlink to the old lib directory.

jrafanie:
We'll remove this symlink eventually, but this will get things working quicker.


abellotti:
LGTM :cake:

PR #8, opened 2015-07-08
Centos71 Support

abellotti:
Moving Appliance to:
- CentOS 7.1   (systemd/filewalld/journald goodness)
- Apache 2.4


kickstarts/base.ks.erb
```diff
@@ -91,14 +89,14 @@ net-snmp
 net-snmp-libs
 net-snmp-utils
 nfs-utils
-nodejs010-nodejs                 # used for asset pipeline asset compilation
+nodejs                           # used for asset pipeline asset compilation - From epel
```
jrafanie: We should remove the loading of the nodejs010 scl here:  https://github.com/ManageIQ/manageiq-appliance/blob/5dc2c53ddc18aca703f62f4b0f4b9cb3e92ac9f9/LINK/etc/default/evm#L30

kbrock:
this is very exciting

PR #9, opened 2015-07-08
Remove the temporary imagefactory meta and body files every few days.

jrafanie:
Please review @abellotti @brandondunne

### DIFF
--- a/bin/clean_imagefactory_storage.sh
+++ b/bin/clean_imagefactory_storage.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+days_to_keep_storage=2
+imgfac_storage=/build/storage
+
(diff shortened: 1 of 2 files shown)


abellotti:
let's add that puppy as /build/bin/clean_imagefactory_storage.sh and update the doc so that the cron simply runs that script.  This way I can manually run the script by hand and not needed to setup cron on non-production build VMs.

brandondunne:
:+1: LGTM

kbrock:
:+1: yay clean all things!

jrafanie:
Good idea @abellotti, done.  Please review.

abellotti:
Looks Good. Thanks @jrafanie

PR #12, opened 2015-07-09
dont autospawn appliance_console

kbrock:
This is related to ManageIQ/manageiq-appliance#2 and ManageIQ/manageiq#792

No longer want to use the miqconsole instead of the getty.
Just use the default login.

Database first run no longer depends upon the no longer existant miqconsole service

/cc @abellotti @Fryguy


abellotti:
:+1:

PR #13, opened 2015-07-09
Remove nightly scripts

jrafanie:
These were replaced with cron via:
c453b3e and 11af18a

@abellotti Please review  :scissors:


scripts/vmbuild.rb
```diff
@@ -47,6 +47,7 @@
     end
     $log.info("Checking out reference #{cli_options[:reference]} from repo #{build_repo} ...")
     `git reset --hard`                                    # Drop any local changes
+    `git clean -dxf`                                      # Clean up any local untracked changes
```
abellotti: :+1: Nice Find.

PR #14, opened 2015-07-09
Use relative paths to the "latest" symlink.

abellotti:
Updated the vmbuild.rb script so that "latest" symlinks are created with relative path to the target directories.

This allows the latest links to be traversed from different VM build servers that cross mount the fileshare.

screen shot 2015-07-09 at 1 53 50 pm

@jrafanie please review/merge.

@JPrause this will allow users to go to the kegerator subdirectory that they see off virtlab31 or aab-brewery7 and see and be able to click on the latest symlink.

### DIFF
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -170,21 +170,24 @@ Dir.chdir(IMGFAC_DIR) do
   end
 end
 
-# Only update the symlink for a nightly
+# Only update the latest symlink for a nightly
 unless build_label == "test"
-  link = stream_directory.join("latest")
+  symlink_name = "latest"
+  link = stream_directory.join(symlink_name)
   if File.exist?(link)
     raise "#{link} is not a symlink!" unless File.symlink?(link)
     result = FileUtils.rm(link, :verbose => true)
     $log.info("Deleted symlink: #{result}")
   end
 
-  result = FileUtils.ln_s(destination_directory, link, :verbose => true)
-  $log.info("Created symlink: #{result}")
+  Dir.chdir(stream_directory) do
+    result = FileUtils.ln_s(directory_name, symlink_name, :verbose => true)
+    $log.info("Created symlink: #{result}")
+  end
 
   if cli_options[:fileshare] && FILE_SERVER
-    $log.info "Updating latest symlink on #{FILE_SERVER} ..."
-    ssh_cmd = "cd #{file_rdu_dir_base}; rm -f latest; ln -s #{directory_name} latest"
+    $log.info "Updating #{symlink_name} symlink on #{FILE_SERVER} ..."
+    ssh_cmd = "cd #{file_rdu_dir_base}; rm -f #{symlink_name}; ln -s #{directory_name} #{symlink_name}"
     $log.info `ssh #{FILE_SERVER_ACCOUNT}@#{FILE_SERVER} "#{ssh_cmd}"`
   end
 end


scripts/vmbuild.rb
```diff
@@ -170,21 +170,24 @@ def verify_run(output)
...
+  link = stream_directory.join(symlink_name)
   if File.exist?(link)
     raise "#{link} is not a symlink!" unless File.symlink?(link)
     result = FileUtils.rm(link, :verbose => true)
     $log.info("Deleted symlink: #{result}")
   end
 
-  result = FileUtils.ln_s(destination_directory, link, :verbose => true)
-  $log.info("Created symlink: #{result}")
+  Dir.chdir(stream_directory) do
```
matthewd: You shouldn't need to chdir here; ln doesn't care whether the path exists relative to cwd.

```ruby
result = FileUtils.ln_s(directory_name, link, :verbose => true)
```

kbrock: does it make sense to have all this code in the `Dir.chdir(stream_directory)`, not introduce `symlink_name` and just change `link = 'latest'`?

```ruby
Dir.chdir(stream_directory) do
  link = "latest"
  if File.exist?(link)
    raise "#{link} is not a symlink!" unless File.symlink?(link)
    result = FileUtils.rm(link, :verbose => true)
    $log.info("Deleted symlink: #{result}")
  end

  result = FileUtils.ln_s(directory_name, link, :verbose => true)
  $log.info("Created symlink: #{result}")
end

Comment (kbrock) [inline scripts/vmbuild.rb]: agreed that a chdir is probably not necessary.

File: scripts/vmbuild.rb [hunk 2]
```diff
@@ -170,21 +170,24 @@ def verify_run(output)
...
+    result = FileUtils.ln_s(directory_name, symlink_name, :verbose => true)
```
jrafanie: @abellotti I'm confused, won't stream_directory be "/build/fileshare/upstream"?

Are we going to change the fileshare link on the build vm?

Currently, I think this is still modifying the latest link in the vm's directory because the /build/fileshare links to the kegerator's directory:

`/build/fileshare -> /mnt/builds/kegerator`

Don't we need to modify /mnt/builds/upstream/latest instead of just  /mnt/builds/kegerator/latest?

scripts/vmbuild.rb
abellotti: if we're not in that directory, then we need to send the full path of the source file to ln -s, ending up in latest -> /build/.... path.   we really need latest -> directory_name as not all may have things mounted exactly the same way.

scripts/vmbuild.rb
abellotti: /mnt/builds happens to be the same for all build machines (except virtlab31 which has it mounted on /home/appliances/cfme). so build VMs will put their builds in their relative directories, i.e. /mnt/builds/kegerator/, /mnt/builds/aab-brewery7, i.e. the /build/fileshare for that particular build machine.  so /mnt/builds/upstream/latest does not apply here, really is /mnt/builds/kegerator/upstream/latest.    Nice thing about the changes in this PR, it does not matter how/where things are mounted, as as long as its mounted, you can see/traverse the latest links from any VM build machine. even if it's /home/appliances/cfme/ or anything else for that matter.  I hope this makes sense.

jrafanie: :confused: I think we were solving different problems.  I now understand what you were fixing.

scripts/vmbuild.rb
jrafanie: LGTM, @matthewd @kbrock merge or comment if you have other concerns.

matthewd: > if we're not in that directory, then we need to send the full path of the source file to ln -s

Nope:

```
% irb
irb(main):001:0> require 'fileutils'
=> true
irb(main):002:0> FileUtils.ln_s('b', 'tmp/a')
=> 0
irb(main):003:0> 

% ls -l tmp
lrwxr-xr-x  1 matthew  staff    1 10 Jul 05:52 a -> b
```

jrafanie: Ok, I'll review this and test it... will take a little time to test... opened #15 so I don't forget.

PR #17, opened 2015-07-20
Install wmi on appliance build.

carbonin:
The wmic tool is needed to get running processes from windows machines.
This PR is also needed for https://github.com/ManageIQ/manageiq/pull/3540
https://bugzilla.redhat.com/show_bug.cgi?id=1233815

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -127,6 +127,8 @@ openldap-clients
 
 cloud-init
 
+wmi
+
 <% if @target == "ovirt" %>
 ovirt-guest-agent
 <% elsif @target == "vsphere" %>
@@ -169,8 +171,6 @@ open-vm-tools
 %end
 
 %post --log=/root/anaconda-post.log
-
-yum -y install http://www6.atomicorp.com/channels/atomic/centos/7/x86_64/RPMS/wmi-1.3.14-4.el7.art.x86_64.rpm
 
 exec < /dev/tty3 > /dev/tty3
 chvt 3
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -170,6 +170,8 @@ open-vm-tools
 
 %post --log=/root/anaconda-post.log
 
+yum -y install http://www6.atomicorp.com/channels/atomic/centos/7/x86_64/RPMS/wmi-1.3.14-4.el7.art.x86_64.rpm
+
 exec < /dev/tty3 > /dev/tty3
 chvt 3
 set -x


kbrock:
Is there a reason to add this here vs adding to the rpm itself?

carbonin:
Spoke with @brandondunne about this one.  It is already packaged as an rpm for the downstream appliance, but we added it here to install it during the upstream build as well.

kbrock:
:+1: thanks. Of course you are right here. :shipit:

Fryguy:
@brandondunne Please review.

kickstarts/base.ks.erb
```diff
@@ -157,6 +157,8 @@ open-vm-tools
 
 %post --log=/root/anaconda-post.log
 
+yum -y install http://www6.atomicorp.com/channels/atomic/centos/7/x86_64/RPMS/wmi-1.3.14-4.el7.art.x86_64.rpm
```
brandondunne: It would be better to get this RPM from a non-commercial source (like EPEL), however I can't find it anywhere else.  @carbonin Is this the only place that you found it?  We may want to rebuild this in COPR if we can't find it anywhere else.

carbonin: I wasn't able to find it anywhere else either.  Where do we get the package we use for downstream builds?

abellotti: If you have the .el7.src.rpm for it, you can add it in https://github.com/ManageIQ/srpms, then I can kick off a build and have it available in COPR for us to use (our manageiq-scl for CentOS 7.1 repo already added/created in kickstart on line 19).

carbonin: @abellotti Looks like the package we are using for downstream (el6) works fine on a CentOS 7 system, installed and tested it yesterday. Can we just build that into an el7 rpm? (https://brewweb.devel.redhat.com/buildinfo?buildID=425456)

carbonin:
@abellotti @jvlcek Any update on this PR?  I did notice the last comment on the BZ seems to say this may not get in.

Fryguy:
@carbonin Can you respond to https://github.com/ManageIQ/manageiq-appliance-build/pull/17#discussion_r35569627 and/or work with @abellotti on that?

carbonin:
@Fryguy The last I head @abellotti and @jvlcek were looking into where we got the source for the package downstream and would try to use that as the source seems to be platform independent. My last comment was in reference to https://bugzilla.redhat.com/show_bug.cgi?id=1233815#c7

abellotti:
LGTM!!  wmi now available from our COPR SCL and @carbonin tested with appliance built. Merging !!

jvlcek:
:+1:

PR #18, opened 2015-07-28
No need to chdir to the target directory for creating the symlink.

### LINKED ISSUE #15: Change and test symlinking code to not change directories
See: https://github.com/ManageIQ/manageiq-appliance-build/pull/14#discussion_r34285335

It's not a difficult change, but requires some testing... putting this here to remember to do it.

abellotti:
This resolves #15


abellotti:
@jrafanie there you go.

jrafanie:
:+1:

PR #20, opened 2015-08-03
[auth] Need the realmd package

abellotti:
- This provides the tools for realm discovery and enrollment to assist in sssd configuration against directories such as AD or IdM.


kbrock:
and this is were it all begins

kbrock:
:+1: (my comment without the snark)

PR #21, opened 2015-08-10
[auth] Adding modules needed for Active Directory

abellotti:
- Modules needed for joining Active Directories with realm.

https://trello.com/c/kky5CB6r


kbrock:
Yay. +1 to Active Directory.

PR #22, opened 2015-08-10
removed the centos71-devbuild.sh script

abellotti:
This is no longer needed since the 7.1 work was merged on master.


kbrock:
yay! recent OS FTW

PR #23, opened 2015-08-14
Replaced ext4 with xfs in kickstart commands

carbonin:


Fryguy:
cc @jrafanie @simaishi

matthewd:
xref manageiq/manageiq#3882

jrafanie:
:+1:

PR #28, opened 2015-08-21
[WIP] Remove unused productization by adding --only for building specific targets

jrafanie:
* Remove productization since we don't use it for building images 
  * we can externalize parts of the kickstart for customization in the future
  * the current productization is for values we're not really customizing
* Add CLI option `--only` to build only specific targets (since targets.yml is now gone)
  `--only vsphere ovirt`
* Use Pathnames instead of building paths using Strings
* Add testing for scripts via rspec --init
* Extract target information from vmbuild to the Target class.
* Add Gemfile


jrafanie:
@abellotti I haven't tested this yet but wanted to see what you thought.  I hate targets.yml when all I ever do is run vsphere only.

abellotti:
I like what you did here, is targets.yml not used by downstream ?

other approach for cmd-line option would be --targets=vsphere   instead of --vsphere_only, this allows us to build only let's stay openstack, or allow multiple, i.e. --targets=vsphere,ovirt

scripts/vmbuild.rb
```diff
@@ -63,9 +62,8 @@
 
 $log.info "Using Configuration base directory: #{cfg_base}"
 
-targets_file      = Build::Productization.file_for(cfg_base, "config/targets.yml")
-tdl_file          = Build::Productization.file_for(cfg_base, "config/base.tdl")
-ova_file          = Build::Productization.file_for(cfg_base, "config/ova.json")
+tdl_file = BUILD_BASE.join("config/base.tdl")
```
kbrock: much nicer, thank

jrafanie: :smile: trying to remove the cruft :fire:

jrafanie:
> I like what you did here, is targets.yml not used by downstream ?

Yes, it's only used by upstream, so it's pointless to have this complexity.

> other approach for cmd-line option would be --targets=vsphere instead of --vsphere_only, this allows us to build only let's stay openstack, or allow multiple, i.e. --targets=vsphere,ovirt lack of --targets would be all flavors like what you have in the hash.

Yes, I was considering that.  I can do that if you think it's useful.

abellotti:
yes please if you don't mind.  It'll be more flexible.

On Mon, Aug 24, 2015 at 9:56 AM, Joe Rafaniello <notifications@github.com>
wrote:

> I like what you did here, is targets.yml not used by downstream ?
>
> Yes, it's only used by upstream, so it's pointless to have this complexity.
>
> other approach for cmd-line option would be --targets=vsphere instead of
> --vsphere_only, this allows us to build only let's stay openstack, or allow
> multiple, i.e. --targets=vsphere,ovirt lack of --targets would be all
> flavors like what you have in the hash.
>
> Yes, I was considering that. I can do that if you think it's useful.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ManageIQ/manageiq-appliance-build/pull/28#issuecomment-134212646>
> .
>



-- 
*Alberto Bellotti*
abellott@redhat.com

kbrock:
:+1: flexibility. especially since we added the hash for this work

scripts/cli.rb
```diff
@@ -19,17 +19,19 @@ def parse
...
-        opt :build_url, build_desc, :type => :string,  :default => BUILD_URL, :short => "B"
-        opt :manageiq_url, manageiq_desc, :type => :string,  :default => MANAGEIQ_URL, :short => "M"
+        opt :type,          type_desc,      :type => :string,  :short => "t", :default => DEFAULT_TYPE
+        opt :reference,     git_ref_desc,   :type => :string,  :short => "r", :default => DEFAULT_REF
+        opt :local,         local_desc,     :type => :boolean, :short => "l", :default => false
+        opt :fileshare,     share_desc,     :type => :boolean, :short => "s", :default => true
+        opt :appliance_url, appliance_desc, :type => :string,  :short => "A", :default => APPLIANCE_URL
+        opt :build_url,     build_desc,     :type => :string,  :short => "B", :default => BUILD_URL
+        opt :manageiq_url,  manageiq_desc,  :type => :string,  :short => "M", :default => MANAGEIQ_URL
+        opt :vsphere_only,  vsphere_desc,   :type => :boolean, :short => "V", :default => false
```
Fryguy: would it make sense to just have a `--only` and then give it the keys of the things you want to build?  So, for vsphere only one could do `--only vsphere`.  If you want 2 out of 3 or something you could do `--only vsphere,ovirt`.  trollop supports Array types.

scripts/vmbuild.rb
```diff
@@ -84,16 +82,25 @@ def verify_run(output)
...
-targets_config    = YAML.load_file(targets_file)
-name, directory, targets = targets_config.values_at("name", "directory", "targets")
+
+directory       = "upstream"
+name            = "manageiq"
+
+targets = {
+  "vsphere"   => "vsphere",
+  "ovirt"     => "rhevm",
+  "openstack" => "openstack-kvm"
```
kbrock: Think these are the values to accept for the `--only` CLI option

scripts/vmbuild.rb
```diff
@@ -84,16 +82,25 @@ def verify_run(output)
 directory_name    = "#{year_month_day}_#{hour_minute}"
 timestamp         = "#{year_month_day}#{hour_minute}"
 
-targets_config    = YAML.load_file(targets_file)
-name, directory, targets = targets_config.values_at("name", "directory", "targets")
+
+directory       = "upstream"
+name            = "manageiq"
```
Fryguy: Minor but the far alignment is not necessary

scripts/cli.rb
Fryguy: Oh I just saw the --targets comment from @abellotti .  I didn't realize it was already suggested :)

jrafanie: Yes, it's similar to what @abellotti suggested.  Thanks for all the feedback.  I'll try to get back to this and get it updated for review.

scripts/vmbuild.rb
jrafanie: Thanks for commenting on that.  I now noticed the naming is not so good since it was moved from the targets.yml to here... directory and name are pretty ambiguous in this context.

scripts/cli.rb
jrafanie: > Oh I just saw the --targets comment from @abellotti . I didn't realize it was already suggested :)

Clearly, it's a good idea if two different people thought of it independently.  I secretly did the easiest thing first to see if others thought it was a good idea.  :wink:

jrafanie:
Updated, major changes since last review:
* change vsphere_only to only and accept a list of targets
* extract target information (supported types, imagefactory type names, file extension) to a Build::Target class
* Added basic testing by adding rspec --init, gemfile

jrafanie:
@abellotti @Fryguy @kbrock Please throw :tomato: at this.  I'll test a build with this if you think it looks ok.

scripts/cli.rb
```diff
@@ -1,5 +1,6 @@
 require 'rubygems'
 require 'trollop'
+require_relative 'target'
```
kbrock: could you namespace this (and not use require_relative)?

```ruby
require 'build/target'
```

scripts/cli.rb
```diff
@@ -11,7 +12,7 @@ class Cli
     BUILD_URL     = "https://github.com/ManageIQ/manageiq-appliance-build.git"
     MANAGEIQ_URL  = "https://github.com/ManageIQ/manageiq.git"
 
-    def parse
+    def parse(args = ARGV)
```
kbrock: :+1:

scripts/cli.rb
```diff
@@ -19,17 +20,19 @@ def parse
...
         banner "Usage: build.rb [options]"
 
-        opt :type,        type_desc,     :type => :string,  :default => DEFAULT_TYPE, :short => "t"
-        opt :reference,   git_ref_desc,  :type => :string,  :default => DEFAULT_REF,  :short => "r"
-        opt :local,       local_desc,    :type => :boolean, :default => false,        :short => "l"
-        opt :fileshare,   share_desc,    :type => :boolean, :default => true,         :short => "s"
-        opt :appliance_url, appliance_desc,  :type => :string,  :default => APPLIANCE_URL, :short => "A"
-        opt :build_url, build_desc, :type => :string,  :default => BUILD_URL, :short => "B"
-        opt :manageiq_url, manageiq_desc, :type => :string,  :default => MANAGEIQ_URL, :short => "M"
+        opt :type,          type_desc,      :type => :string,  :short => "t", :default => DEFAULT_TYPE
```
kbrock: odd that we are adding all these `:short` entries when it can be derived.

scripts/Gemfile
```diff
@@ -0,0 +1,4 @@
+source 'https://rubygems.org'
```
kbrock: would this be better served in a `gemspec`?
almost feels like the script directory is becoming a gem

scripts/kickstart_generator.rb
```diff
@@ -13,7 +11,7 @@ class KickstartGenerator
     attr_reader :targets, :puddle, :appliance_checkout, :manageiq_checkout
 
     def initialize(build_base, targets, puddle, appliance_checkout, manageiq_checkout)
-      @build_base         = Pathname.new(build_base)
+      @build_base         = build_base
```
kbrock: I like what you are doing here and the implications throughout the code

scripts/kickstart_generator.rb
```diff
@@ -59,7 +57,7 @@ def write_json(result)
     end
 
     def evaluate_erb
-      ks_file = Productization.file_for(@build_base, "#{KS_DIR}/base.ks.erb")
+      ks_file = @build_base.join("#{KS_DIR}/base.ks.erb")
```
kbrock: not sure what I like better:

```ruby
@build_base.join(KS_DIR, "base.ks.erb")
```

scripts/spec/cli_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+require 'cli'
+
+describe Build::Cli do
+  context "#parse" do
+    it "only (default)" do
+      expect(described_class.new.parse(%w()).options[:only]).to match_array %w(vsphere ovirt openstack)
```
kbrock: kudos

scripts/spec/cli_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+    end
+
+    it "only vsphere and ovirt" do
+      expect(described_class.new.parse(%w(-o vsphere ovirt)).options[:only]).to match_array %w(vsphere ovirt)
```
kbrock: neat / odd / scratching my head. I didn't expect that input to map to that output value.

How does it know this is not `options = {"-o" => "vsphere"} ; ARGV=["ovirt"]`

I thought it would be `-o vsphere,ovirt` or `-o vsphere -o ovirt`

scripts/cli.rb
kbrock: this is back to the gem question...

scripts/cli.rb
```diff
@@ -1,5 +1,6 @@
 require 'rubygems'
```
kbrock: do we want to keep this?

scripts/cli.rb
```diff
@@ -1,5 +1,6 @@
...
 require 'trollop'
+require_relative 'target'
 
 module Build
   class Cli
```
kbrock: Thanks for adding the `Cli` class and parse. I've been trying to follow this trend with `ApplianceConsole::Cli` and `FixAuth::Cli`

kbrock:
This is looking great.
Besides asking about pulling out into own gem (which can be done in a future PR / never), I'm good

scripts/spec/cli_spec.rb
kbrock: ok, the difference between `:multi => true` and `:type ==> :strings`. confusing. Who manages this darn trollop gem anyway?

jrafanie: Yeah, tell me about it :trollface:

jrafanie:
@kbrock I'll try to address your concerns in #36... Closing this one since productization is still something we want to use for the kickstart but not for the vmbuild.rb specifically.

PR #19, opened 2015-07-31
Upgrade postgresql92 to rh-postgresql94.

jrafanie:
Change postgresql92 SCL references to rh-postgresql94.

https://trello.com/c/aLn9eAj1

Related PRs:
ManageIQ/manageiq#3676
ManageIQ/manageiq-appliance#9

PR #29, opened 2015-08-21
Postgresql94 cleanup

jrafanie:
I noticed some cleanup while testing pg94 in #19.

The only directories within COPY are now etc, var, and usr (we no longer have files we don't want to copy in opt) and we want to copy them all so let's blindly copy all files below COPY as that is most people would expect.


jrafanie:
@abellotti @kbrock Please review/merge.  Thanks!

PR #30, opened 2015-08-21
External Authentication - Adding support for LDAP

abellotti:
- Adding the openldap-clients package (ldapsearch, etc.), these tools are commonly
used when debugging manual configurations of OpenLDAP, certificates from directory
servers and such before diving into sssd.conf.

https://trello.com/c/SIOuHAMk/87-auth-3-support-ldap-authentication-via-sssd


jrafanie:
:+1:

PR #31, opened 2015-08-27
v2_key no longer exists by default

jrafanie:
Depends on: https://github.com/ManageIQ/manageiq/pull/4082

@kbrock Please review, @Fryguy helped me on both 4082 and this one.


kbrock:
good job @jrafanie

PR #32, opened 2015-08-28
Remove extra shared memory needed for %post running PG setup.

jrafanie:
We used to run initdb and migrate the database as part of the kickstart
%post and had to configure normal shared memory but since we now setup
the database on first deployment boot, we shouldn't need this.

cc @carbonin @abellotti


abellotti:
:+1:

kbrock:
:+1:

PR #33, opened 2015-08-28
Appliance repo shouldn't be in www root.

jrafanie:
manage-appliance repo shouldn't be in www root.

Related PRs:
* https://github.com/ManageIQ/manageiq-appliance-build/pull/33 [ clone repo in new location ] 
* https://github.com/ManageIQ/manageiq-appliance/pull/12  [set environment variable for new location]
* https://github.com/ManageIQ/manageiq/pull/4105 [use environment variable above]

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -200,7 +200,7 @@ firewall-offline-cmd --zone=manageiq --add-service=postgresql
 
 www_root="/var/www"
 app_root="$www_root/miq"
-appliance_root="$www_root/manageiq-appliance"
+appliance_root="/opt/manageiq/manageiq-appliance"
 
 rm -rf $app_root
 


jrafanie:
@simaishi @carbonin @Fryguy please review.

kbrock:
I like this change, but should we create a symbolic link?

my muscle memory and many docs say `/var/www/miq/...`

abellotti:
I like it too, but it feels like we're not all there. I vote for moving both rails app and appliance files there.
/opt/manageiq/manageiq/ 
/opt/manageiq/manageiq-appliance/

And the appropriate links from /var/www/miq/ to minimize field shock, and updates to restorecon's.

Thoughts ?

jrafanie:
> I like it too, but it feels like we're not all there. I vote for moving both rails app and appliance files there.
> /opt/manageiq/manageiq/ 
> /opt/manageiq/manageiq-appliance/
> 
> And the appropriate links from /var/www/miq/ to minimize field shock, and updates to restorecon's.
> 
> Thoughts ?

We chickened out. @Fryguy, @simaishi, and I thought that moving manageiq out of vmdb was too risky.
There are a few things that use realpath code in ruby, like bundler binstubs, so a symlink was not really tried.  I guess we could try it, I'm not sure how much work it was.  manageiq-appliance doesn't have the same risk since we just extracted it a few weeks ago.

jrafanie:
Also, keep in mind that ruby's require will require the same file twice (once via symlink, once via real path) so you could get already initialized constant warnings and worse if we traverse both real and symlink to require a file.

jrafanie:
See: https://github.com/ManageIQ/manageiq/pull/3451

jrafanie:
I hope to remove the symlinks for old lib and old system soon too.

jrafanie:
@kbrock @abellotti what do you think?  I'm concerned with using symlinks due to it's limitations with require and code really needs to learn to work without symlinks.  With that said, we don't intend to move MIQ code out of /var/www/miq/vmdb yet.  This is only for manageiq-appliance code, former system directory.

kbrock:
we'll need to update selinux directories if we move. symlink won't protect us

:+1: moving the appliance code. I don't think our code links into there (didn't check)

PR #34, opened 2015-09-01
Remove cloud-init config changes into cfme-setup.sh

carbonin:
These commands don't belong in the kickstart.  Moved to cfme-setup.sh in https://github.com/ManageIQ/manageiq-appliance/pull/13


kbrock:
:+1:

TANGENT: I'll restate again. I find it very confusing that manageiq-appliance-build and appliance are separate repos. I understand the reasoning behind why we did that, but I think we may want to ask if we should merge back again.

Again, nick, this affects you more than me.

PR #35, opened 2015-09-01
Separate kickstart into partials

carbonin:
This will allow us to simplify the base kickstart file by adding partials that are more easily edited.


abellotti:
LGTM. @jrafanie ?

kbrock:
If this is getting complex, then good idea.
If we are going to create multiple base kickstarts, then good idea.

But if is is just a hundred or so line file with a partial that is only used by 1 file, not sure what it buys us.
staying as close to the metal would be nice

carbonin:
@kbrock I think the idea is to allow us to move the productization stuff into partials, rather than having to copy and maintain changes to the entire base kickstart downstream.  Plus the complexity point.  @jrafanie may be able to give more details about background.

carbonin:
For now I'm also looking for other pieces that we may want to break out in the PR in addition to the filesystem bits.

Fryguy:
:+1: Looks good.  I think let's upWIP, get it merged, and then do separate PRs for other parts.

kickstarts/base.ks.erb
```diff
@@ -36,21 +36,7 @@ device mptspi
...
-part /var/www/miq_tmp --ondrive=vda                     --size=10240 --fstype=xfs --fsoptions="rw,noatime,nobarrier"
-part pv.2             --ondrive=vda                     --size=10240 --grow
-
-volgroup vg_system    --pesize=4096 pv.1
-logvol /              --name=lv_os   --vgname=vg_system --size=10240 --fstype=xfs
-logvol swap           --name=lv_swap --vgname=vg_system --size=4096  --grow
-
-volgroup vg_data      --pesize=4096 pv.2
-logvol /var/opt/rh/rh-postgresql94/lib/pgsql/data --name=lv_pg --vgname=vg_data --size=8192 --grow --fstype=xfs
+<%= render_partial "disk_layout" %>
```
jrafanie: What are the thoughts on order within the kickstart... I think everything in the main section of the kickstart are not order dependent in terms of "disk layout, repo definitions, etc."  I believe that leaves only %post where order matters.  I ask because the filename doesn't indicate order.  

Do we want to have this name main_disk_layout?

jrafanie:
> If this is getting complex, then good idea.

Yes, the packages section, coming up soon, is many many lines and makes the kickstart hard to read.

>If we are going to create multiple base kickstarts, then good idea.

Yes, that's part of the intent.  If we need to try a newer centos or rhel, it's easy enough to do just the partials that needs to be changed.

>But if is is just a hundred or so line file with a partial that is only used by 1 file, not sure what it buys us. staying as close to the metal would be nice

Agreed, although splitting this out into sections may help us organize it better in the first place.  But, we do have an intent to use this for other base kickstarts.

kickstarts/base.ks.erb
carbonin: Ah right, nearly forgot that we talked about order.  So then we would preface any other partials within the "main" section with main_?  If we want to just separate the section files from each other, we could also do it with a directory structure by creating a partials/main and making the kickstart line `<%= render_partial "main/disk_layout" %>`.

kbrock:
:+1: thanks for background. this looks great :shipit:

kickstarts/base.ks.erb
jrafanie: `<%= render_partial "main/disk_layout" %>.`

I :heart: that.  If you're making changes to the %post section, you can basically ignore all of the main section/directory.  You very rarely care about it.  Having the directories to separate the sections filters out the partials you don't care about.

jrafanie: Can you just verify that all areas within the main are really order independent?  If so, we can name them as you have... in post, we'd have to name them in order so we know the order of the templates somehow.

carbonin: According to https://docs.fedoraproject.org/en-US/Fedora/21/html/Installation_Guide/chap-kickstart-installations.html 
>Sections must be specified in order. Items within the sections do not have to be in a specific order unless otherwise specified.
So it seems like we are good for the main section unless @abellotti says otherwise.

PR #36, opened 2015-09-01
Add cli for targets

jrafanie:
Replaces #28 

We don't use productization for vmbuild.rb so this PR removes productization for vmbuild.rb, thereby deleting the targets.yml, and instead exposes them as CLI options.


scripts/target.rb
```diff
@@ -0,0 +1,42 @@
+module Build
+  class Target
+    TARGET_MAPPING = {
```
kbrock: Grouping the target mapping by the name may simplify things
```ruby
MAPPINGS = {
  'vsphere' => {
    'target' => 'vsphere',
    'extenstion' => 'ova'
  }
}

def initialize(name)
  @name = name = name.to_s
  @mapping = MAPPING[name]
end

def imagefactory_type
  @mapping['target']
end
```

Heck, or even:

```ruby
  class Target < OStruct(:name, :imagefactory_type, :file_extension)
  end

# this may not access the map 100%
def self.get_mapping(name)
  return Target.new(name, TARGET_MAPPING[name], FILE_EXTENSION[name])
end
```

kbrock:
ugh, could spend a month in here "fixing it"

@Fryguy is this good enough. ship it?

jrafanie:
@kbrock I have a few more things I want to fix, I'll mark it wip and get it cleaned up today/tomorrow.  I want to get it in so @carbonin can write tests.

kbrock:
the only suggestion I have is to use a factor: `Build.get_target(name)` or `Target.find_by_name(name)`

`Target` seems like a simple struct with hardcoded values. With a struct, this all could be done w/ 10 lines of code.

jrafanie:
@kbrock @Fryguy Ok, rebased, and used a struct to simplify the Target class a bit.  Should I do a gemspec instead of a gemfile here? Anything else?

Fryguy:
It's not a gem, so no on the gemspec.

Fryguy:
My only comment is to use `eq` not `eql`.  `eql` maps to Ruby `eql?` and `eq` maps to Ruby `==`

PR #38, opened 2015-09-04
Fix precompile_assets post-reroot and productization move.

jrafanie:
Note, it assumes you checkout manageiq as sister directories to manageiq-appliance-build.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,10 +1,11 @@
-require_relative 'productization'
+require_relative 'scripts/productization'
+require 'pathname'
 
 namespace :build do
 
   module FilePaths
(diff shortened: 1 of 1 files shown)


kbrock:
I hadn't realized that `__dir__` was lowercase vs uppercase `__FILE__`. (then again I don't use the dir method because I can never find it)

all in all, this just renames the `vmdb` directory to `manageiq`.

:+1:

jrafanie:
> I hadn't realized that __dir__ was lowercase vs uppercase __FILE__. (then again I don't use the dir method because I can never find it)

Yeah, I bet there was a bike shed over the name of that one.  I guess `__DIR__` was conflicting with something else.  Either way, `__dir__` is a useful replacement over `File.dirname(__FILE__)`

PR #39, opened 2015-09-08
Add new --upload option

abellotti:
we don't want to upload nightlies as release builds from all VM build machines,  only from the
real build machine, i.e. if run from the cron via bin/nightly-build.sh.  Also updated README.md for the
required fog gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,11 @@
 
   ```
   yum install ruby
+  yum install ruby-devel
+  yum install zlib-devel
+
   gem install trollop
+  gem install fog
   ```
 
   * For enabling copying to SSH file server, define the following in Root's .bashrc (optional)
--- a/scripts/cli.rb
+++ b/scripts/cli.rb
@@ -19,7 +19,7 @@ module Build
       appliance_desc = "Repo URL containing appliance scripts and configs(COPY/LINK/TEMPLATE)"
       build_desc     = "Repo URL containing the build config and kickstart"
       manageiq_desc  = "Repo URL containing the main manageiq code"
-      upload_desc    = "Upload nightly build to release builds"
+      upload_desc    = "Upload appliance builds to the website"
 
       @options = Trollop.options do
         banner "Usage: build.rb [options]"
--- a/scripts/cli.rb
+++ b/scripts/cli.rb
@@ -12,24 +12,26 @@ module Build
     MANAGEIQ_URL  = "https://github.com/ManageIQ/manageiq.git"
 
     def parse
-      git_ref_desc  = "provide a git reference such as a branch or tag"
-      type_desc     = "build type: nightly, test, a named yum repository"
-      local_desc    = "Use local config and kickstart for build"
-      share_desc    = "Copy builds to file share"
+      git_ref_desc   = "provide a git reference such as a branch or tag"
+      type_desc      = "build type: nightly, test, a named yum repository"
+      local_desc     = "Use local config and kickstart for build"
+      share_desc     = "Copy builds to file share"
       appliance_desc = "Repo URL containing appliance scripts and configs(COPY/LINK/TEMPLATE)"
       build_desc     = "Repo URL containing the build config and kickstart"
       manageiq_desc  = "Repo URL containing the main manageiq code"
+      upload_desc    = "Upload nightly build to release builds"
 
       @options = Trollop.options do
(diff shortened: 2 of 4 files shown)


abellotti:
@jrafanie please review.

kbrock:
this looks nice.

[soapbox]can we stop reindenting stuff? you added a few lines and it made it look like you changed 1/2 the script - git blame is no longer helpful either[/soapbox]

scripts/cli.rb
```diff
@@ -12,24 +12,26 @@ class Cli
...
-      local_desc    = "Use local config and kickstart for build"
-      share_desc    = "Copy builds to file share"
+      git_ref_desc   = "provide a git reference such as a branch or tag"
+      type_desc      = "build type: nightly, test, a named yum repository"
+      local_desc     = "Use local config and kickstart for build"
+      share_desc     = "Copy builds to file share"
       appliance_desc = "Repo URL containing appliance scripts and configs(COPY/LINK/TEMPLATE)"
       build_desc     = "Repo URL containing the build config and kickstart"
       manageiq_desc  = "Repo URL containing the main manageiq code"
+      upload_desc    = "Upload nightly build to release builds"
```
jrafanie: Upload appliance builds to the website

README.md
```diff
@@ -220,7 +220,11 @@
 
   ```
   yum install ruby
+  yum install ruby-devel
+  yum install zlib-devel
+
   gem install trollop
+  gem install fog
```
jrafanie: What do these changes have to do with the --upload option?

scripts/cli.rb
```diff
@@ -12,24 +12,26 @@ class Cli
...
-        opt :build_url, build_desc, :type => :string,  :default => BUILD_URL, :short => "B"
-        opt :manageiq_url, manageiq_desc, :type => :string,  :default => MANAGEIQ_URL, :short => "M"
+        opt :type,          type_desc,      :type => :string,  :default => DEFAULT_TYPE,  :short => "t"
+        opt :reference,     git_ref_desc,   :type => :string,  :default => DEFAULT_REF,   :short => "r"
+        opt :local,         local_desc,     :type => :boolean, :default => false,         :short => "l"
+        opt :fileshare,     share_desc,     :type => :boolean, :default => true,          :short => "s"
+        opt :appliance_url, appliance_desc, :type => :string,  :default => APPLIANCE_URL, :short => "A"
+        opt :build_url,     build_desc,     :type => :string,  :default => BUILD_URL,     :short => "B"
+        opt :manageiq_url,  manageiq_desc,  :type => :string,  :default => MANAGEIQ_URL,  :short => "M"
+        opt :upload,        upload_desc,    :type => :boolean, :default => false,         :short => "u"
```
jrafanie: Let's add this option in one commit and align with padding (to allow a few spaces for each section) so we don't have to align these each time we add a new line with 1 or 2 more characters.  Thoughts?

jrafanie:
Nice @abellotti, good call, I forgot this would affect test builds.  Thanks :clap:

README.md
abellotti: uploader.rb requires fog even if we don't use it.

jrafanie: Thanks!  I forgot I made those changes to get it to work. What about `libxml2-devel` and `libxslt-devel`?  I think something in the fog dependency tre uses nokogiri.

abellotti: right, the yum install ruby-devel/zlib-devel were needed when gem install fog was rebuilding nokogiri.

Fryguy: Why not just have a Gemfile?

scripts/cli.rb
Fryguy: Nearly all of the types can be removed, btw.  It is inferred from the default.

Fryguy: Same goes for the short values.  They are automatically assigned for a large number of these cases.

Fryguy:
Looks good to me :+1:  @jrafanie Can we merge anyway ignoring the alignment stuff?

README.md
jrafanie: Yeah, I need to get back to that, I started down that route here  where I needed a project skeleton (gemfile/gemspec/rspec) #36, haven't finished that one up yet

jrafanie:
Looks good, merging.

PR #41, opened 2015-09-08
Create more kickstart partials

carbonin:
Move more things out of the base kickstart file and into partials.


kickstarts/partials/post/source_setup.ks.erb
```diff
@@ -0,0 +1,56 @@
...
+        done
+      popd
+    done
+  popd
+
+  ln -vs $appliance_root/LINK/.toprc /.toprc
+
+  #copy all files/directories below COPY
+  cp -vr COPY/* /
+popd
```
jrafanie: This section from the `pushd $appliance_root`  down to here belongs in a partial called :hankey: 

Just kidding, maybe this should be in it's own partial called `partials/post/link_copy_system_files.ks.erb`?

kickstarts/partials/post/source_setup.ks.erb
```diff
@@ -0,0 +1,56 @@
...
+  git checkout <%= @manageiq_checkout.branch %>
+  git reset --hard <%= @manageiq_checkout.commit_sha %>
+popd
+
+# Symlink extracted repo (manageiq-appliance) to old /var/www/miq/system location.
+ln -vs $appliance_root $app_root/system
+
+# Symlink old lib directory to gems/pending.
+ln -vs $app_root/vmdb/gems/pending $app_root/lib
+
```
jrafanie: These top 25 lines should be by themselves... note, we will be dropping these `ln -vs` lines very shortly since we no longer need and shouldn't symlink these directories anymore.

kickstarts/base.ks.erb
```diff
@@ -167,169 +50,21 @@ set -x
...
-  git reset --hard <%= @manageiq_checkout.commit_sha %>
-popd
-
-# Symlink extracted repo (manageiq-appliance) to old /var/www/miq/system location.
-ln -vs $appliance_root $app_root/system
+<%= render_partial "post/firewalld" %>
 
-# Symlink old lib directory to gems/pending.
-ln -vs $app_root/vmdb/gems/pending $app_root/lib
+<%= render_partial "post/source_setup" %>
```
jrafanie: We definitely need numbers in the names of these things as it will be hard to tell what files in the post section go in what order.  What do you think?

Fryguy: The order is defined by how they are listed in the base.ks.erb

kickstarts/partials/post/db_init.ks.erb
```diff
@@ -0,0 +1,27 @@
+# Create a script to initialize appliance on first boot
+cat > /bin/appliance-initialize.sh <<EOF
+#!/bin/sh
+[[ -d /var/opt/rh/rh-postgresql94/lib/pgsql/data/base ]] && exit 0
+[[ -s /etc/default/evm ]] && source /etc/default/evm
+echo "Initializing Appliance, please wait ..." > /dev/tty1
+appliance_console_cli --region 0 --internal --password smartvm
+EOF
```
Fryguy: This really needs to be ripped out of kickstart (not in this PR, just saying)

kickstarts/partials/post/repos.ks.erb
```diff
@@ -0,0 +1,7 @@
+# post install repos
+# yum update uses these repos to update and reinstall packages.
+# Please also add to "build time repos" around line 20
```
Fryguy: This comment no longer makes sense.   Can you update to refer to the corresponding partial?

kickstarts/base.ks.erb
jrafanie: Yeah, we lose that context when it's extracted.  Maybe it's not a problem?  I guess we can see what it looks like after it's done.

kickstarts/partials/post/db_init.ks.erb
jrafanie: Yes, it's nice that it's separated out because it's easier to grok all by itself.

kickstarts/partials/post/source_setup.ks.erb
carbonin: Yeah, I kept them together because they all use the variables defined at the top of the file.  It doesn't make a difference functionally because it is just text at this point, but it might be tough to edit the file if we are using variables defined in another partial.

kickstarts/partials/post/source_setup.ks.erb
carbonin: Same issue as on line 25.  I wouldn't want to move the code into a partial other than where the variables are defined.

PR #42, opened 2015-09-08
Changes needed for vmdb reroot and productization move

simaishi:
@Fryguy @jrafanie 
Fix file_for argument and chdir for tar command
Change tar command not to use include and update exclude list, needed due to vmdb reroot

@Fryguy For 'exclude', verified empty 'tmp' and 'log' directories are included in tar file.


Rakefile
```diff
@@ -43,8 +43,7 @@ namespace :build do
 
   desc "Builds a tarball."
   task :tar => [:version_files, :build_file, :precompile_assets] do
-    include_file = Build::Productization.file_for("config/tarball/include")
```
jrafanie: I guess this means you get to also delete the include file?  :scissors:  :+1: Please also delete that file if it's not needed.

Rakefile
```diff
@@ -56,7 +55,7 @@ namespace :build do
     transform = RUBY_PLATFORM =~ /darwin/ ? "-s " : "--transform s"
     transform << "',^,#{tar_basename}/,'"
 
-    `tar -C .. #{transform} -T #{include_file} -X #{exclude_file} -hcvzf #{tarball}`
+    `tar -C ../manageiq #{transform} -X #{exclude_file} -hcvzf #{tarball} .`
```
jrafanie: what is the final `.` adding to the tarball?

config/tarball/exclude
```diff
@@ -1,20 +1,13 @@
...
-*vmdb/config/database.yml
-*vmdb/certs/v2_key
-*vmdb/log/*.log
-*vmdb/tmp/miq_temp
-*lib/disk/modules/ruby1.9.3
+*.git
+Gemfile.lock*
+.bundle
+bin
+certs/v2_key.dev
```
jrafanie: Nice, yes, we don't want this key.  :+1:

Rakefile
simaishi: everything from manageiq minus what's in exclude file

jrafanie:
@simaishi I like it, but please delete the include file if we don't need it anymore :fire:

We may have to tweak this include list based on what we see in the tarball on various upstream versions going forward.

config/tarball/exclude
```diff
@@ -1,20 +1,13 @@
...
-*vmdb/bin
-*vmdb/config/database.yml
-*vmdb/certs/v2_key
-*vmdb/log/*.log
-*vmdb/tmp/miq_temp
-*lib/disk/modules/ruby1.9.3
+*.git
+Gemfile.lock*
+.bundle
+bin
```
matthewd: We want to keep this now

Rakefile
simaishi: Include file removed.

config/tarball/exclude
Fryguy: We were just discussing this...for right now we are going to exclude it because the gemset builder will put it right back in with a symlink.  We have to change the gemset builder to handle this differently, but for now we are excluding it.

Fryguy:
Merging...I will add a trello task to review putting back the bin exclude and using it properly.

PR #44, opened 2015-09-09
Add script spec tests

carbonin:
Add some spec tests for productization and kickstart generator classes.

### DIFF
--- a/scripts/spec/kickstart_generator_spec.rb
+++ b/scripts/spec/kickstart_generator_spec.rb
@@ -1,0 +1,43 @@
+require 'spec_helper'
+require 'kickstart_generator'
+
+describe Build::KickstartGenerator do
+  describe "#run" do
+    def build_base
+      data_file_path("kickstart_generator")
+    end
+
+    def generated
+      "#{build_base}/kickstarts/generated"
+    end
+
+    def ks_text
+      File.read("#{generated}/base.ks")
+    end
+
+    after(:each) do
+      FileUtils.rm_f(Dir.glob("#{generated}/*"))
+    end
+
+    before(:each) do
+      @generator = described_class.new(build_base, ["target"], "puddle", "app_checkout", "miq_checkout")
+      @generator.run(nil)
+    end
+
+    it "writes ks file" do
+      expect(ks_text).to include("base file text")
+    end
+
+    it "renders partial files" do
+      expect(ks_text).to include("first partial file text")
+    end
+
+    it "renders partials in partial files" do
+      expect(ks_text).to include("second partial file text")
+    end
+
+    it "renders partial files in subdirectories" do
+      expect(ks_text).to include("subdir partial text")
+    end
+  end
+end
--- a/scripts/spec/spec_helper.rb
+++ b/scripts/spec/spec_helper.rb
@@ -97,3 +97,7 @@ end
 
 require 'pathname'
 $LOAD_PATH << Pathname.new(__dir__).join("..")
+
+def data_file_path(to)
+  File.expand_path(to, File.join(File.dirname(__FILE__), "data"))
+end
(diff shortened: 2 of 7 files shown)


scripts/spec/kickstart_generator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      File.read("#{generated}/base.ks")
+    end
+
+    after(:each) do
+      FileUtils.rm_f(Dir.glob("#{generated}/*"))
+    end
+
+    before(:each) do
+      @generator = described_class.new(build_base, ["target"], "puddle", "app_checkout", "miq_checkout")
+      @generator.run(nil)
```
jrafanie: This is fine, maybe `run` should return an object you can call GeneratedKickstart#config to return a String of the ini (readable) format and #json to return a json String which is what imagefactory needs to use.

You could then do:

```ruby
expect(@generator.run(nil).config).to include("base file text")
```

It's fine the way it is but maybe that makes sense.

scripts/spec/spec_helper.rb
```diff
@@ -97,3 +97,7 @@
 
 require 'pathname'
 $LOAD_PATH << Pathname.new(__dir__).join("..")
+
+def data_file_path(to)
+  File.expand_path(to, File.join(File.dirname(__FILE__), "data"))
```
Fryguy: This can be simplified to `File.join(File.expand_path(to, __dir__), data)`

jrafanie:
@carbonin Can you make any change and rebase this PR to see if travis runs on it?    Thanks!

scripts/spec/kickstart_generator_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper'
+require 'kickstart_generator'
+
+describe Build::KickstartGenerator do
+  describe "#run" do
+    def build_base
+      data_file_path("kickstart_generator")
```
kbrock: You probably want to use [mktmpdir](http://ruby-doc.org/stdlib-2.0.0/libdoc/tmpdir/rdoc/Dir.html) instead.

scripts/spec/kickstart_generator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    def generated
+      "#{build_base}/kickstarts/generated"
+    end
+
+    def ks_text
+      File.read("#{generated}/base.ks")
+    end
+
+    after(:each) do
+      FileUtils.rm_f(Dir.glob("#{generated}/*"))
```
kbrock: see above note. (putting dirname into a let makes for easy access at cleanup time)

scripts/spec/kickstart_generator_spec.rb
kbrock: 1. use a let for the generator
2. `run` returns `self`, maybe the same for some of the other methods in there, too.

kbrock:
1. Feels like specs for scripts should live in `spec/scripts/`
2. Feels like this code belongs in `lib` I like conventions from `rails`/ `bundle gem` where the script/exe is ~5 lines of code and the meat is over in lib.

scripts/spec/kickstart_generator_spec.rb
carbonin: For what?  I'm using the directory structure that would be in place during a real build run which is recreated in the `data/kickstart_generator` directory.  In this tree are the base.ks.erb and partials I'm using to test; assuming I'm not also going to generate those files from within the test I can't be creating the build base here as a temp directory.

carbonin:
@kbrock I'm fine with those ideas, but they seem a bit out of scope for this PR.

jrafanie:
@kbrock we need to change the scripts directory to lib, move the spec directory up to the root and properly namespace the repo.  It doesn't have to be a gem but it should have a layout that looks more like a gem.

PR #46, opened 2015-09-10
Remove now that qpid is unsupported! :heart:

jrafanie:
Depends on: https://github.com/ManageIQ/manageiq/pull/4291

cc @blomquisg


jrafanie:
cc @simaishi

kbrock:
so much joy :sob:

PR #48, opened 2015-09-11
Fix uninitialized constant error for Target class

carbonin:
@jrafanie 
Fix for:
/build/scripts/vmbuild.rb:92:in `block in <main>`: uninitialized constant Target (NameError)
	from /build/scripts/vmbuild.rb:92:in `collect`
	from /build/scripts/vmbuild.rb:92:in `<main>`


jrafanie:
:+1:

PR #51, opened 2015-09-14
Created filesystems for STIG compliance

carbonin:
Created separate file systems for:
`/tmp`
`/home`
`/var/log/audit`
`/var/log`
`/var`
`/`

https://bugzilla.redhat.com/show_bug.cgi?id=1180768
https://trello.com/c/B10qwxfk

### DIFF
--- a/kickstarts/partials/main/disk_layout.ks.erb
+++ b/kickstarts/partials/main/disk_layout.ks.erb
@@ -8,10 +8,10 @@ part /var/www/miq_tmp --ondrive=vda                     --size=10240 --fstype=xf
 part pv.2             --ondrive=vda                     --size=10240 --grow
 
 volgroup vg_system    --pesize=4096 pv.1
+logvol /              --name=lv_os             --vgname=vg_system --size=4680 --fstype=xfs
+logvol /var           --name=lv_var            --vgname=vg_system --size=2048 --fstype=xfs
+logvol /var/log       --name=lv_var_log        --vgname=vg_system --size=1024 --fstype=xfs
+logvol /var/log/audit --name=lv_var_log_audit  --vgname=vg_system --size=512  --fstype=xfs
 logvol /tmp           --name=lv_tmp            --vgname=vg_system --size=1024 --fstype=xfs
 logvol /home          --name=lv_home           --vgname=vg_system --size=1024 --fstype=xfs
-logvol /var/log/audit --name=lv_var_log_audit  --vgname=vg_system --size=512  --fstype=xfs
-logvol /var/log       --name=lv_var_log        --vgname=vg_system --size=1024 --fstype=xfs
-logvol /var           --name=lv_var            --vgname=vg_system --size=2048 --fstype=xfs
-logvol /              --name=lv_os             --vgname=vg_system --size=4680 --fstype=xfs
 logvol swap           --name=lv_swap           --vgname=vg_system --size=4096 --grow
--- a/kickstarts/partials/main/disk_layout.ks.erb
+++ b/kickstarts/partials/main/disk_layout.ks.erb
@@ -8,5 +8,10 @@ part /var/www/miq_tmp --ondrive=vda                     --size=10240 --fstype=xf
 part pv.2             --ondrive=vda                     --size=10240 --grow
 
 volgroup vg_system    --pesize=4096 pv.1
-logvol /              --name=lv_os   --vgname=vg_system --size=10240 --fstype=xfs
-logvol swap           --name=lv_swap --vgname=vg_system --size=4096  --grow
+logvol /tmp           --name=lv_tmp            --vgname=vg_system --size=1024 --fstype=xfs
+logvol /home          --name=lv_home           --vgname=vg_system --size=1024 --fstype=xfs
+logvol /var/log/audit --name=lv_var_log_audit  --vgname=vg_system --size=512  --fstype=xfs
+logvol /var/log       --name=lv_var_log        --vgname=vg_system --size=1024 --fstype=xfs
+logvol /var           --name=lv_var            --vgname=vg_system --size=2048 --fstype=xfs
+logvol /              --name=lv_os             --vgname=vg_system --size=4680 --fstype=xfs
+logvol swap           --name=lv_swap           --vgname=vg_system --size=4096 --grow


carbonin:
@jrafanie @abellotti and others please review sizes

kbrock:
This is great.

1. Do we want to minimize root and have a bigger mount for `/opt/`?
2. Do we want `/var/log` to be bigger?

carbonin:
@kbrock 
I only added the file systems that were explicitly required for the STIG standard, so if we want a new file system for `/opt` that might be a different conversation.

Also I think `/var/log` probably isn't going to be as much of a space hog as `/var` as that's where our application logs are going.  Is there something specifically that could be taking up a ton of space in `/var/log`?  Also we can easily set the behavior of journald to only use a specific amount of space if we need to (http://www.freedesktop.org/software/systemd/man/journald.conf.html).  I think it will be the only thing using significant space in `/var/log`

kbrock:
@carbonin ping jared. He used to talk about logs filling up.
Also, we have a process for log rotating that checks if disk is full. That may need to change if all of var is not on the same disk.

carbonin:
@kbrock Looks like we should be fine with the log rotation script.  It checks based on filesystem that the log file is on. (https://github.com/ManageIQ/manageiq-appliance/blob/master/logrotate_free_space_check.sh#L4)

kbrock:
:+1: then looks good to me

kickstarts/partials/main/disk_layout.ks.erb
```diff
@@ -8,5 +8,10 @@ part /var/www/miq_tmp --ondrive=vda                     --size=10240 --fstype=xf
...
 
 volgroup vg_system    --pesize=4096 pv.1
-logvol /              --name=lv_os   --vgname=vg_system --size=10240 --fstype=xfs
-logvol swap           --name=lv_swap --vgname=vg_system --size=4096  --grow
+logvol /tmp           --name=lv_tmp            --vgname=vg_system --size=1024 --fstype=xfs
+logvol /home          --name=lv_home           --vgname=vg_system --size=1024 --fstype=xfs
+logvol /var/log/audit --name=lv_var_log_audit  --vgname=vg_system --size=512  --fstype=xfs
+logvol /var/log       --name=lv_var_log        --vgname=vg_system --size=1024 --fstype=xfs
+logvol /var           --name=lv_var            --vgname=vg_system --size=2048 --fstype=xfs
+logvol /              --name=lv_os             --vgname=vg_system --size=4680 --fstype=xfs
```
jrafanie: @carbonin nitpicky stuff here but since I believe only the `swap` line below needs to be last due to the --grow option, can we order the above lines by importance to our application?

Maybe something like this:

```
/    # os
/var  # app
/var/log  # os logs
/var/log/audit
/tmp
/home
```

jrafanie:
@carbonin these values look good to me at least as a first try so we can get some feedback as to what sizes they really should be.

kickstarts/partials/main/disk_layout.ks.erb
```diff
@@ -8,5 +8,10 @@ part /var/www/miq_tmp --ondrive=vda                     --size=10240 --fstype=xf
 part pv.2             --ondrive=vda                     --size=10240 --grow
 
 volgroup vg_system    --pesize=4096 pv.1
-logvol /              --name=lv_os   --vgname=vg_system --size=10240 --fstype=xfs
-logvol swap           --name=lv_swap --vgname=vg_system --size=4096  --grow
+logvol /              --name=lv_os             --vgname=vg_system --size=4680 --fstype=xfs
+logvol /var           --name=lv_var            --vgname=vg_system --size=2048 --fstype=xfs
+logvol /var/log       --name=lv_var_log        --vgname=vg_system --size=1024 --fstype=xfs
+logvol /var/log/audit --name=lv_var_log_audit  --vgname=vg_system --size=512  --fstype=xfs
+logvol /tmp           --name=lv_tmp            --vgname=vg_system --size=1024 --fstype=xfs
```
brandondunne: @roliveri When we fleece Openstack images, do we download those to `/tmp`?

roliveri: It depends. Upstream, it places the temp file where ever Ruby's `Tempfile` determines it should go.
Downstream has the option to add an additional disk for temp usage. Then, it'll mount the disk under `/var/www/miq_tmp` and download the images to that filesystem. If the extra disk isn't added, it's the same as upstream.

carbonin: I thought we used https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L7 for fleecing Openstack upstream?

brandondunne: Tempfile on linux points to `/tmp`, either way is seems likely that we will run out of disk space on one of those filesystems when scanning Openstack.

kickstarts/partials/main/disk_layout.ks.erb
```diff
@@ -8,5 +8,10 @@ part /var/www/miq_tmp --ondrive=vda                     --size=10240 --fstype=xf
 part pv.2             --ondrive=vda                     --size=10240 --grow
 
 volgroup vg_system    --pesize=4096 pv.1
-logvol /              --name=lv_os   --vgname=vg_system --size=10240 --fstype=xfs
-logvol swap           --name=lv_swap --vgname=vg_system --size=4096  --grow
+logvol /              --name=lv_os             --vgname=vg_system --size=4680 --fstype=xfs
```
brandondunne: Is this too small to contain the database?  I think we ship with a 40GB disk, right?

carbonin: @brandondunne The database is on a separate disk https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/db_fs.ks.erb

kickstarts/partials/main/disk_layout.ks.erb
jrafanie: I was under the impression we use `/var/www/miq_tmp` if it exists otherwise we use the temp/Tempfile location.  If so, I think we're ok here since we still have the existing ~10 GB temp disk mounted at that location.  I haven't looked at that code recently though.

chessbyte:
@carbonin What is the status of this PR?
/cc @gtanzillo @jrafanie @bdunne @roliveri @Fryguy @kbrock @abellotti

kbrock:
FYI - I've come across a number of QE people who run out of disk space during the install. they tend to have 9GB for their database partition - which is not enough for most of their test databases.

carbonin:
@kbrock This shouldn't be touching anything to do with databases as we put that on a separate partition (https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/db_fs.ks.erb).
@chessbyte I think the only issue we could have is if @roliveri is correct and we always use `Tempfile` for openstack fleecing.  If that is the case we should probably remove https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L7 and give the space to `/tmp`

jrafanie:
> @kbrock This shouldn't be touching anything to do with databases as we put that on a separate partition (https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/db_fs.ks.erb).
> @chessbyte I think the only issue we could have is if @roliveri is correct and we always use Tempfile for openstack fleecing. If that is the case we should probably remove https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L7 and give the space to /tmp

Based on the info here, I'm :+1: on giving the space where Tempfile will use it on linux and remove the mounted miq_tmp location.

If we want to adjust some of that space to give more to the database (unchanged in this PR), we can do that here or in a followup PR.

kbrock:
Is this mergeable? I have no further concerns.

PR #60, opened 2015-09-22
Used variable for appliance repo path instead of system symlink

carbonin:
@jrafanie @abellotti @simaishi


jrafanie:
Nice @carbonin!  Sorry, didn't see this until now.  Thanks @chessbyte

PR #63, opened 2015-09-29
Add self service UI build

simaishi:
Updated downstream build wiki to include instruction on how to install node, gulp and bower on build machine.

@Fryguy

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -43,8 +43,15 @@ namespace :build do
     Dir.chdir(__dir__)
   end
 
+  task :build_selfservice_ui do
+    Dir.chdir(File.join(__dir__, '../manageiq/spa_ui/self_service'))
+    puts `npm install`
+    puts `git clean -xdf`  # cleanup temp files
+    Dir.chdir(__dir__)
+  end
+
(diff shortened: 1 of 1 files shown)


kbrock:
@simaishi the build sure is progressing. thanks
@Fryguy almost feels like self service ui is a separate repo...

Rakefile
```diff
@@ -43,8 +43,15 @@ namespace :build do
     Dir.chdir(__dir__)
   end
 
+  task :build_selfservice_ui do
+    Dir.chdir(File.join(__dir__, '../manageiq/spa_ui/self_service'))
+    puts `npm install`
+    puts `git clean -xdf`  # cleanup temp files
+    Dir.chdir(__dir__)
```
Fryguy: I'll merge for now, but when we come back to refactor this, you can use the block form of Dir.chdir instead, which brings you back when you're done:

```ruby
puts Dir.pwd # "/Users/jfrey/"
Dir.chdir("/") do
  puts Dir.pwd # "/"
end
puts Dir.pwd # "/Users/jfrey/"
```

PR #66, opened 2015-10-13
Stopped and disabled NetworkManager

carbonin:
We don't need it and it makes configuring static IPs by editing files by hand less reliable.

### DIFF
--- a/kickstarts/partials/post/systemd.ks.erb
+++ b/kickstarts/partials/post/systemd.ks.erb
@@ -14,3 +14,6 @@ ln -sf /dev/null /etc/systemd/system/ctrl-alt-del.target
 # Create the journal directory to enable persistant logging
 mkdir /var/log/journal
 systemctl restart systemd-journald
+
+systemctl stop NetworkManager
+systemctl disable NetworkManager


carbonin:
@abellotti What do you think?

kbrock:
While this is an appliance, it would be nice if this were configured similar to most server environments. If 75% of all systemd servers have this enabled, may be nice to enable it.

But this is outside my core, so go for it.

carbonin:
@kbrock I think it is actually the opposite, most server environments disable it in favor of configuring the networking manually.

Additionally, we're not removing it entirely so if someone wanted to configure the network themselves with `nmcli` or one of the other tools provided by NetworkManager that is still possible.

abellotti:
@carbonin  @kbrock

 Well, disabling network manager broke something.

We no longer have the updated /etc/issue and /etc/issue.net

aka:

```
Welcome to the Virtual Appliance

You can browse to http://x.x.x.x or http://fqdn
```

This was generated for us via /etc/NetworkManager/dispatcher.d/99-fix-issue

carbonin:
@abellotti It doesn't look like that file is doing anything particularly special.  Why does it need to be managed by NetworkManager?  Couldn't we come up with another way to have to get run on startup, a systemd service maybe?

abellotti:
that's fine, as long as it runs.

btw that file is missing ` ` around the sed  in the PRODUCT=... line somehow


On Thu, Oct 29, 2015 at 1:37 PM, Nick Carboni <notifications@github.com>
wrote:

> @abellotti <https://github.com/abellotti> It doesn't look like that file
> is doing anything particularly special. Why does it need to be managed by
> NetworkManager? Couldn't we come up with another way to have to get run on
> startup, a systemd service maybe?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ManageIQ/manageiq-appliance-build/pull/66#issuecomment-152260013>
> .
>



-- 
*Alberto Bellotti*
abellott@redhat.com

abellotti:
Issue entered https://github.com/ManageIQ/manageiq-appliance-build/issues/73

PR #68, opened 2015-10-20
Compile sti loader and keep yml in tar

simaishi:
Not sure if there is a better/cleaner way to handle exclusion from exclusion for tar creation (exclude everything in tmp except for sti_loader.yml)..

@jrafanie @Fryguy

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -40,7 +40,11 @@ namespace :build do
   task :precompile_assets do
     Dir.chdir(File.join(__dir__, '..', 'manageiq')) do
       puts `bundle exec rake evm:compile_assets`
+    end
+  end
 
+  task :precompile_sti_loader do
+    Dir.chdir(File.join(__dir__, '..', 'manageiq')) do
       # compile_sti_loader fails without database.yml - copy as temporary solution
       FileUtils.cp("config/database.pg.yml", "config/database.yml")
       puts `bundle exec rake evm:compile_sti_loader`
@@ -56,7 +60,7 @@ namespace :build do
   end
 
   desc "Builds a tarball."
-  task :tar => [:version_files, :build_file, :precompile_assets, :build_selfservice_ui] do
+  task :tar => [:version_files, :build_file, :precompile_assets, :precompile_sti_loader, :build_selfservice_ui] do
     exclude_file = Build::Productization.file_for((__dir__), "config/tarball/exclude")
     pkg_path     = Pathname.new(__dir__).join("pkg")
     FileUtils.mkdir_p(pkg_path)
--- a/Rakefile
+++ b/Rakefile
@@ -38,16 +38,21 @@ namespace :build do
   end
 
   task :precompile_assets do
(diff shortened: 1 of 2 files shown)


config/tarball/exclude
```diff
@@ -1,7 +1,6 @@
 *.DS_Store
 */spec
 */test
-*/tmp/*
```
jrafanie: Is this going to include tmp/pids/*.pid if it exists?

Rakefile
```diff
@@ -65,7 +70,8 @@ namespace :build do
     transform = RUBY_PLATFORM =~ /darwin/ ? "-s " : "--transform s"
     transform << "',^,#{tar_basename}/,'"
 
-    `tar -C ../manageiq #{transform} -X #{exclude_file} -hcvzf #{tarball} .`
+    # Everything from */tmp/* should be excluded, except for tmp/cache/sti_loader.yml
+    `tar -C ../manageiq #{transform} --exclude-tag='cache/sti_loader.yml' -X #{exclude_file} -hcvzf #{tarball} .`
```
jrafanie: I'm not sure what exclude-tag does... I think we want to exclude all of tmp but allow tmp/cache/sti_loader.yml if that's possible....

config/tarball/exclude
simaishi: yes, it will remove tmp/pids/*.pid if it exists.

Rakefile
simaishi: That's what I'm trying to achieve - only keep tmp/cache/sti_loader.yml.

Using exclude-tag, it will remove everything from the directory that contains the file specified, with the exception of the actual file.

Since what's specified in exclude-tag (cache/sti_loader.yml) is under 'tmp', this removes everything from tmp but keeping tmp/cache/sti_loader.yml.

jrafanie:
Looks good to me @simaishi... can you confirm this worked for you?

Rakefile
kbrock: @matthewd alternatively, we could have the loader look for a file in a different directory.

Had you considered that we may pre-compile this file and ship it?

simaishi:
@jrafanie Yes, this worked fine locally.  The only file left in tmp was the yml file.

Rakefile
simaishi: @kbrock This PR is for pre-compiling the file and including in the tar file so appliance will have it on the machine when shipped.  Not sure if you're asking something different..?

kbrock: @simaishi this is great. ship it
Just making sure @matthewd understands how this is being used.

putting files into /tmp are typically temp files.
But it looks like we are precompiling assets and precompiling dependencies.
So just want that to be known as we further evolve the design.

again. not a concern. :shipit:

jrafanie: @kbrock I agree, while this is a temp file, it's a different type of temporary file,  we might want to reconsider where this file is stored if it becomes more of a headache if for example, we need to add another temporary file to this directory.

jrafanie:
:ship:

PR #72, opened 2015-10-29
Put the DEVICE line in ifcfg files if missing

carbonin:
Ensures that we will have valid interface configuration at first boot.

@simaishi @bdunne @abellotti @gtanzillo

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -67,6 +67,14 @@ $appliance_root/cfme-setup.sh
 
 <%= render_partial "post/db_init" %>
 
+# make sure we have the device name in all ifcfg-* files
+ls /etc/sysconfig/network-scripts/ifcfg-* | while read FILE
+do
+  # parse the device name from FILE
+  DEVICE=${FILE##*-}
+  grep -q 'DEVICE=' $FILE && echo "DEVICE=$DEVICE" >> $FILE
+done
+
 # Let's rebuild the ramfs with with base scsi drivers we need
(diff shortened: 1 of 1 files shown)


kickstarts/base.ks.erb
```diff
@@ -67,6 +67,13 @@ $appliance_root/cfme-setup.sh
 
 <%= render_partial "post/db_init" %>
 
+# make sure we have the device name in all ifcfg-* files
+ls /etc/sysconfig/network-scripts/ifcfg-* | while read FILE
+do
+  DEVICE=`echo $FILE | cut -d'/' -f5 | cut -d'-' -f2`
+  [[ -z "`grep 'DEVICE=' $FILE`" ]] && echo "DEVICE=$DEVICE" >> $FILE
```
kbrock: looks good

Protip 1:

I think you could use the return status from grep instead of `-z`. something like:

```bash
[[ grep -q 'DEVICE=' $FILE ]] || echo "DEVICE=$DEVICE" >> $FILE
```

protip 2:

I think you can get the device name using bash `DEVICE=${FILE##*-}`

kbrock:
this looks good to me

@carbonin thanks for edits. I was just sharing more than requesting fixes...

carbonin:
No I did like the ideas.  I didn't know about the syntax for parsing the file path.  That was a nice find.  I really didn't like having to call so many commands to grab the device.

gtanzillo:
:+1: Looks good to me too!

bdunne:
:+1: LGTM

PR #74, opened 2015-10-29
Add vhd support for Hyper-V/SVCMM

simaishi:
Tested by building/deploying hyperv and vsphere images with this change.

Note: Depending on the version of qemu-img on the machine, the final image could be the same size as the disk size.  CentOS 7.1 produced 40GB image, Fedora22 produced 3.8GB image.  Use -o to limit type of images to build as needed.

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -78,7 +78,7 @@ done
 # Let's rebuild the ramfs with with base scsi drivers we need
 kversion=$(rpm -q kernel --qf '%{version}-%{release}.%{arch}\n')
 ramfsfile="/boot/initramfs-$kversion.img"
-/sbin/dracut --force --add-drivers "mptbase mptscsih mptspi" $ramfsfile $kversion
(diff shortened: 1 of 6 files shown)


abellotti:
Until kegerator's OS is updated to support better vhd compression with qemu-img, we also need an option added to bin/nightly-build.sh so that it does not generate the vhd. either a new option like *--skip hyperv* or we need to specify all other flavors with *--only vsphere,ovirt,openstack*

simaishi:
Is it acceptable if I change nightly-build.sh to pass any remaining arguments (after shifting local, fg, etc.) to vmbuild.sh as build options?

Or do we want to support specifying/limiting image flavors only?

simaishi:
Updated to exclude hyperv by default.

abellotti:
couldn't tell from code, does "--only hyperv" work ?

simaishi:
Yes, that's how I built hyperv for testing.  There is no change to how "-o" works if you specify specific image(s) to build, unless you specify "all".

abellotti:
:+1: LGTM !!

PR #77, opened 2015-11-06
Enable ports 5900-5999 for HTML5 console

carbonin:
Removed the `firewall` command in favor of opening
ports in the post section.

The kickstart `firewall` command was adding the ports
to the default zone which is "public" at the time the
main section runs.

We later create the "manageiq" zone and set that to the
default causing the ports to still be closed in the active
zone.

https://bugzilla.redhat.com/show_bug.cgi?id=1278432


carbonin:
@abellotti @gtanzillo @martinpovolny

gtanzillo:
:+1: LGTM

kickstarts/partials/post/firewalld.ks.erb
```diff
@@ -15,3 +15,10 @@ firewall-offline-cmd --zone=manageiq --add-service=http
 firewall-offline-cmd --zone=manageiq --add-service=https
 firewall-offline-cmd --zone=manageiq --add-service=ssh
 firewall-offline-cmd --zone=manageiq --add-service=postgresql
+firewall-offline-cmd --zone=manageiq --add-port=22/tcp
+firewall-offline-cmd --zone=manageiq --add-port=80/tcp
+firewall-offline-cmd --zone=manageiq --add-port=80/udp
+firewall-offline-cmd --zone=manageiq --add-port=443/tcp
+firewall-offline-cmd --zone=manageiq --add-port=443/udp
+firewall-offline-cmd --zone=manageiq --add-port=5432/tcp
+firewall-offline-cmd --zone=manageiq --add-port=5900-5999/tcp
```
kbrock: this is so much nicer than listing out every port.

carbonin: And it works :smile:

PR #78, opened 2015-11-06
WIP: Add base_image retry support.

jrafanie:
The appliance build code has been getting hung when it's creating the base image during the rpm installation phase, it complains when communicating with yum repository or it's contents.  Usually, the problem is that it says the pg94 repo doesn't contain the postgresql94-postgresql-server package.

The existing code detects that the base image creation fails to properly create an imagefactory image uuid and then exits without first killing the currently running image (via `virsh destroy N`).  This leads to the first failed build to hang forever until we kill it manually and start over.

The attached code uses the existing check for the uuid and instead of exiting, it attempts to blindly retry the base image creation after X minutes.  It will wait `attempt * attempt * 60`  seconds, (1 minute, 4 minutes, 9 minutes, etc.) between retries, until it's tried 5 times and it finally cleans up and fails the build.

The problem with this retry logic is it doesn't know the reason we failed to create the base image as that's hidden away in the guest vm that's sitting at the failure screen.  I suppose if we run out of disk space, this retry logic could make things worse by trying it again and again when we have no room to build the image.

Thoughts?
:confused:

### DIFF
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -72,13 +72,52 @@ ova_file = BUILD_BASE.join("config/ova.json")
 $log.info "Using inputs: puddle: #{puddle}, build_label: #{build_label}"
 $log.info "              tdl_file: #{tdl_file}, ova_file: #{ova_file}."
 
-def verify_run(output)
+def check_uuid(output)
   if output =~ /UUID: (.*)/
     Regexp.last_match[1]
   else
-    $log.error("Could not find UUID.")
-    exit 1
+    $log.error("Could not find UUID.  Output from last command: #{output}")
+    false
   end
+end
+
+# TODO: Move these things to a class
+def clean_factory_build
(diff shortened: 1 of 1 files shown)


carbonin:
I think the biggest concern with disk space is are the raw disk files in `/build/storage` and it looks like you are removing those between retries, so if that works I don't think disk space should be an issue.

carbonin:
If we really are worried about disk usage we could put `/build/storage` on a separate filesystem so we will at least be able to clean up if/when it does run out of space.

jrafanie:
> If we really are worried about disk usage we could put /build/storage on a separate filesystem so we will at least be able to clean up if/when it does run out of space.

Yeah, I don't know how much of a concern it is... the concern is we could have "other" failures that we're blindly retrying.  I mean, I think this would work... it automates what I do by hand... Granted, I don't know if `virsh destroy X` is synchronous... If any part of it is async, I suppose the script could try to delete the .meta and .body file before the virsh destroy command completes.  I don't know.

PR #79, opened 2015-11-17
Add a new line to /etc/ssh/sshd_config

carbonin:
The current scap-security-guide remediations for the sshd_config
file don't check if there is a new line and append fixes anyway.
This will cause the first fix to share a line with whatever is
at the end of the file.

https://bugzilla.redhat.com/show_bug.cgi?id=1219230
https://bugzilla.redhat.com/show_bug.cgi?id=1219227
https://bugzilla.redhat.com/show_bug.cgi?id=1219231
https://bugzilla.redhat.com/show_bug.cgi?id=1219229
https://bugzilla.redhat.com/show_bug.cgi?id=1219228

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -73,6 +73,9 @@ do
   grep -q 'DEVICE=' $FILE || echo "DEVICE=$DEVICE" >> $FILE
 done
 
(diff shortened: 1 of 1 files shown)


gtanzillo:
:+1:  LGTM

PR #80, opened 2015-12-04
SCL repos like to move on us, use copr instead.

jrafanie:


jrafanie:
This is the code change we made to produce the capablanca-1 build.

PR #83, opened 2015-12-22
Add chrony package to include list

carbonin:
We need this because we fail to start the app without syncing the time settings.

### DIFF
--- a/kickstarts/partials/packages/includes.ks.erb
+++ b/kickstarts/partials/packages/includes.ks.erb
@@ -19,6 +19,7 @@ openssl-devel
 readline-devel
 zlib-devel
 
+chrony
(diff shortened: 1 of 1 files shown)


gtanzillo:
:+1:

jrafanie:
@carbonin should we link to the docs with this really useful info: [The chrony suite is installed by default on some versions of Red Hat Enterprise Linux 7](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/sect-Using_chrony.html#sect-Installing_chrony)

jrafanie:
Note, I couldn't find the same doc in centos7 land but the doc above is public and found via google.

PR #84, opened 2016-01-19
smem provides PSS and USS which are useful for forked processes

jrafanie:
Related:  https://github.com/ManageIQ/manageiq/pull/3593


jrafanie:
cc @akrzos @Fryguy

PR #90, opened 2016-02-17
Start memcached before anything loads rails and tries to connect.

jrafanie:
For rails 5, we switched to rails' MemCacheStore [1], a subclass of rack's
store that tries to connect to memcached when loading rails. The prior store,
DalliStore, did not connect on initialize. We need to start memcached here,
before anything loads Rails.  See also [2].

[1] https://github.com/ManageIQ/manageiq/pull/6704
[2] petergoldstein/dalli #423

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -58,6 +58,8 @@ unset DEBUG
 
 <%= render_partial "post/bundler" %>
 
+<%= render_partial "post/start_memcached_hack" %>
+
 # appliance_root="/opt/manageiq/manageiq-appliance" -- in post/source_setup partial
 $appliance_root/cfme-setup.sh
 
--- a/kickstarts/partials/post/start_memcached_hack.ks.erb
+++ b/kickstarts/partials/post/start_memcached_hack.ks.erb
@@ -1,0 +1,9 @@
+# For rails 5, we switched to rails' MemCacheStore [1], a subclass of rack's
+# store that tries to connect to memcached when loading rails. The prior store,
+# DalliStore, did not connect on initialize. We need to start memcached here,
+# before anything loads Rails.  See also [2].
+#
+# [1] https://github.com/ManageIQ/manageiq/pull/6704
+# [2] https://github.com/petergoldstein/dalli/pull/423
+systemctl enable memcached
+systemctl start memcached


jrafanie:
cc @matthewd @bdunne @carbonin 

FYI @simaishi, MIQ on rails 5 will currently try to connect to memcached 127.0.0.1:12111 when you try to compile assets.  This is one hack that may fix the current build error in nightlies:

```
+ /opt/manageiq/manageiq-appliance/cfme-setup.sh
DEPRECATION WARNING: `serve_static_files` is deprecated and will be removed in Rails 5.1.
Please use `public_file_server.enabled = false` instead.
 (called from block in <top (required)> at /var/www/miq/vmdb/config/environments/production.rb:14)
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from singleton class at /var/www/miq/vmdb/lib/extensions/ar_process.rb:6)
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from <class:Preloader> at /var/www/miq/vmdb/lib/extensions/ar_virtual.rb:283)
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from <module:FinderMethods> at /var/www/miq/vmdb/lib/extensions/ar_virtual.rb:310)
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from <module:Calculations> at /var/www/miq/vmdb/lib/extensions/ar_virtual.rb:317)
rake aborted!
Dalli::RingError: No server available
/var/www/miq/vmdb/config/environment.rb:5:in `<top (required)>'
/var/www/miq/vmdb/lib/tasks/evm.rake:46:in `block (3 levels) in <top (required)>'
/var/www/miq/vmdb/lib/tasks/evm_rake_helper.rb:7:in `with_dummy_database_url_configuration'
/var/www/miq/vmdb/lib/tasks/evm.rake:45:in `block (2 levels) in <top (required)>'
Tasks: TOP => environment
(See full trace by running task with --trace)
```

kbrock:
you know, it tries to connect to the relational database as well, but we bypass it.
wonder if the ENV hack is a good solution for memcache too.

that being said - lets merge this to get to green... but may want to keep that in the back of our mind

:shipit:

carbonin:
I'm pretty sure this won't help. The issue is that the post section is executed in the chroot environment. Currently any `systemctl start` or `systemctl stop` calls seem to be getting ignored.

For example we try to restart `systemd-journald` so that some changes take effect but the anaconda-post.log lists the following output:
```
+ systemctl restart systemd-journald
Running in chroot, ignoring request.
```

jrafanie:
Thanks @carbonin

PR #91, opened 2016-02-18
Remove service start and stop from post section

carbonin:
These commands don't do anything as the post is executed
in a chroot environment.

This prevents errors in anaconda-post.log of the form:
```
  + systemctl restart systemd-journald
  Running in chroot, ignoring request.
```


carbonin:
@jrafanie @simaishi

jrafanie:
:+1:

PR #92, opened 2016-02-18
Use memory store since loading rails 5 fails to connect to memcached.

jrafanie:
We don't need a session store in anaconda so force Rails to use memory store
since rails 5 mem_cache_store tries and fails to connect to a not running
memcached.

Depends on: https://github.com/ManageIQ/manageiq/pull/6751


jrafanie:
Replaces #90

kbrock:
:+1: yay

Fryguy:
:( I hate more hacks...would love a to find a way to not have this at all

PR #93, opened 2016-02-22
Add memcached to Requires= and After= in appliance-initialize

carbonin:
This is needed because the environment rake task will fail if the memcache server is not running.

Motivated by the same issue as https://github.com/ManageIQ/manageiq-appliance-build/pull/92


carbonin:
@jrafanie

Issue #95, opened 2016-02-26
Test builds from a repo branch awkward 


jvlcek:
Currently if a developer wants to test a change pushed to a branch of one of the 4 repos
used by the build that branch must be available on all 4 repos even if changes being
tested are only made to one.

The fix should _optionally_ allow a different **_branch_** name, referred to as a **_reference_** in the build scripts, to be specified for each of the repo URL used by the build scrips.

jvlcek:
@miq-bot assign @jvlcek

jvlcek:
fix merged

PR #96, opened 2016-02-26
Optionally allow a different branch for each repo

jvlcek:
https://github.com/ManageIQ/manageiq-appliance-build/issues/95

This PR will make building development image to test proposed changes simpler.
This will be done by adding optional command line arguments where specific branch names
can be specified for each of the repos.

No default behavior is changing.

### DIFF
--- a/scripts/cli.rb
+++ b/scripts/cli.rb
@@ -27,17 +27,20 @@ module Build
 
       @options = Trollop.options(args) do
         banner "Usage: build.rb [options]"
-
-        opt :type,          type_desc,      :type => :string,  :short => "t", :default => DEFAULT_TYPE
-        opt :reference,     git_ref_desc,   :type => :string,  :short => "r", :default => DEFAULT_REF
-        opt :local,         local_desc,     :type => :boolean, :short => "l", :default => false
-        opt :fileshare,     share_desc,     :type => :boolean, :short => "s", :default => true
-        opt :manageiq_url,  manageiq_desc,  :type => :string,  :short => "M", :default => MANAGEIQ_URL
+        opt :appliance_ref, git_ref_desc,   :type => :string,  :short => "a", :default => DEFAULT_REF
         opt :appliance_url, appliance_desc, :type => :string,  :short => "A", :default => APPLIANCE_URL
+        opt :build_ref,     git_ref_desc,   :type => :string,  :short => "b", :default => DEFAULT_REF
         opt :build_url,     build_desc,     :type => :string,  :short => "B", :default => BUILD_URL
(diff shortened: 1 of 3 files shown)


scripts/cli.rb
```diff
@@ -29,13 +29,16 @@ def parse(args = ARGV)
         banner "Usage: build.rb [options]"
 
         opt :type,          type_desc,      :type => :string,  :short => "t", :default => DEFAULT_TYPE
-        opt :reference,     git_ref_desc,   :type => :string,  :short => "r", :default => DEFAULT_REF
         opt :local,         local_desc,     :type => :boolean, :short => "l", :default => false
-        opt :fileshare,     share_desc,     :type => :boolean, :short => "s", :default => true
+        opt :fileshare,     share_desc,     :type => :boolean, :short => "f", :default => true
```
chessbyte: should we alphabetize this list of options for clarity?

jvlcek:
@abellotti Please review.
I have used these changes to successfully build a test image with the command of the form:


`ruby /build/scripts/vmbuild.rb --type nightly --only openstack -b branch_name -B https://github.com/jvlcek/manageiq-appliance-build.git`

scripts/cli.rb
jvlcek: @chessbyte :+1: Willdo

jvlcek:
I'm working on fixing the travis failure now.

abellotti:
Hi Joe, looks good.

l think we still need to support the -r option for compatibility, we have scripts that make use of that option,  in the code if -r is specified, set to -m, -a and -b (if not specified).  That will also make it a nice shortcut if we really want all three to be the same.

Thanks for the quick turnaround on this !!

jvlcek:
@abellotti Thank you. Yes I had reviewed this with @simaishi  and she also suggested maintaining the -r flag. Which I have now done.

scripts/cli.rb
```diff
@@ -46,11 +49,19 @@ def parse(args = ARGV)
...
+      # --reference overrides all other reference arguments
+      unless options[:reference].to_s.empty?
+        options[:manageiq_ref]  = options[:reference]
+        options[:appliance_ref] = options[:reference]
+        options[:build_ref]     = options[:reference]
+        options[:ssui_ref]      = options[:reference]
+      end
+
+      Trollop.die(:manageiq_ref, git_ref_desc) if options[:manageiq_ref].to_s.empty?
+      options[:manageiq_ref] = options[:manageiq_ref].to_s.strip
```
kbrock: I think you want to move this to 53.5 - before using reference to set all the other references

jvlcek: :+1: Thank you @kbrock

kbrock:
LGTM :+1:

PR #98, opened 2016-02-29
Provide a mechanism for developers to test kickstart ERBs changes

### LINKED ISSUE #97: Need a mechanism to test updates to kickstarts and kickstart partials
Currently there is no easy way for developers to test updates to the kickstart.ERB or the
kickstart partials without running a full build.

A simple mechanism is needed to allow a developer to ensure the generated kickstarts
are correct when testing changes to the kickstart.ERBs and partials.

jvlcek:
Fixes #97

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 /pkg
 /productization
+kickstarts/generated/base.ks
+kickstarts/generated/*.json
+kickstarts/generated/*.cfg
+scripts/Gemfile.dev.rb
--- a/bin/kickstart_build.rb
+++ b/bin/kickstart_build.rb
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+
+# This script is for development testing. It will generate the kickstarts from the
+# ../kickstarts/base.ks.erb and ../kickstarts/partials/*
+
+require 'logger'
+
+require_relative '../scripts/cli'
+require_relative '../scripts/git_checkout'
+require_relative '../scripts/kickstart_generator'
+
+$log = Logger.new(STDOUT)
+
+options = Build::Cli.parse.options
+
+Build::KickstartGenerator.new(
+  File.expand_path("..", __dir__),
+  options[:only],
+  nil,
+  Build::GitCheckout.new(:remote => options[:manageiq_url],  :ref => options[:manageiq_ref]),
+  Build::GitCheckout.new(:remote => options[:appliance_url], :ref => options[:appliance_ref]),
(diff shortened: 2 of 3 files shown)


jvlcek:
This PR is only for a development tool/script that will allow developers to test kickstart ERB changes directly from a manageiq-appliance-build checkout.

To test changes to a kickstart ERB a developer would clone their fork of the manageiq-appliance-build repository, make the proposed changes to the kickstart ERB and/or partials, run the script proposed by this PR and manually evaluate the output generated in <manageiq-appliance-build repo directory>/kickstarts/generated

e.g.:
`
clone https://github.com/my/manageiq-appliance-build.git
cd manageiq-appliance-build
vim kickstarts/base.ks.erb # making proposed updates
cd scripts
bundle exec ruby kickstart_generator.rb
cat ../kickstarts/generated/*
 `

jvlcek:
@carbonin Please review.

carbonin:
I think this might be better somewhere other than the source file itself.
Maybe make it a rake task or a ruby script in the `bin` directory.

What do you think @abellotti @jrafanie ?

jvlcek:
@carbonin Thank you for the reply. Can you please elaborate on "why" you think moving this out of the source file would be an improvement.

abellotti:
Yeah, let's go with a seperate script, that would allow running it from anywhere. How about bin/kickstart_build.sh

carbonin:
@jvlcek I think that leaving that in the same file doesn't distinguish clearly enough that it is for dev/test purposes rather than a step in the actual build process.  Plus moving it out removes the filename hack that you had to put in to not let it run when being required.

jvlcek:
@carbonin OK I can see that.
@abellotti Thank you for the input.
I'll move it to it's own script in bin/kickstart_build.rb (.rb not .sh)

Thanks!

.gitignore
```diff
@@ -1,2 +1,6 @@
 /pkg
 /productization
+kickstarts/generated/base.ks
+kickstarts/generated/*.json
+kickstarts/generated/*.cfg
```
carbonin: Should probably also add `scripts/Gemfile.dev.rb`

jvlcek: Good point @carbonin! I had missed it because I have `Gemfile.dev.rb` in my $HOME/.gitignore
:+1:

carbonin:
LGTM :+1:

kbrock:
:+1: nice job @jvlcek

jrafanie:
Nice @jvlcek @carbonin @abellotti, looks good.

PR #101, opened 2016-02-29
Build hyperv by default

simaishi:
The instruction has been updated to put newer version of qemu-img to get reasonable size .vhd image (thanks @abellotti )


abellotti:
:+1:

PR #103, opened 2016-03-03
Make the root filesystem a size that makes sense

carbonin:
I think 4680 was a typo of 4608 (4.5 GiB)

@jrafanie Because it was bothering me and @simaishi


jrafanie:
> @jrafanie Because it was bothering me and @simaishi

That's awesome!  I'm imagining both of you thinking "what number is that..."

kbrock:
Make sure you update the change log to let them know that they have 72 less bytes to work with.

(ok. seriously - please don't)

PR #110, opened 2016-03-15
Removed services duplicated by ports in firewall config

### LINKED ISSUE #108: Duplicate ports in firewall config?
https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/firewalld.ks.erb

@carbonin @abellotti Do we need ssh, http, https, postgres listed multiple times in the firewall config?

carbonin:
Fixes #108 

@bdunne @abellotti

### DIFF
--- a/kickstarts/partials/post/firewalld.ks.erb
+++ b/kickstarts/partials/post/firewalld.ks.erb
@@ -15,9 +15,3 @@ firewall-offline-cmd --zone=manageiq --add-service=http
 firewall-offline-cmd --zone=manageiq --add-service=https
 firewall-offline-cmd --zone=manageiq --add-service=ssh
 firewall-offline-cmd --zone=manageiq --add-service=postgresql
-firewall-offline-cmd --zone=manageiq --add-port=22/tcp
-firewall-offline-cmd --zone=manageiq --add-port=80/tcp
-firewall-offline-cmd --zone=manageiq --add-port=80/udp
-firewall-offline-cmd --zone=manageiq --add-port=443/tcp
-firewall-offline-cmd --zone=manageiq --add-port=443/udp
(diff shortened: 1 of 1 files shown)


bdunne:
I'd prefer to see the port references go.  Having the named services clearly describes what is open and what service is using it in my opinion.

bdunne:
cc @jrafanie @Fryguy

carbonin:
The services only cover the tcp protocol on the listed ports so I wouldn't be able to just use the services. I could list the services then additionally add the same port with udp if that would be better.

bdunne:
I don't think we actually need the UDP port, do we?

carbonin:
No idea :sweat_smile:

jrafanie:
> I don't think we actually need the UDP port, do we?

I hope we had a commit message that said why we used it.  I remember needing it for something, but I don't remember

bdunne:
Apache should only be using TCP

carbonin:
Just FYI the ports enabled by a particular service are defined in files in `/usr/lib/firewalld/services/`

jrafanie:
I'd be good with removing UDP from the allowed ports if we run it on an appliance and can browse the UI and hit some rest api endpoints. :fire:

carbonin:
Okay, I'll run a build for this branch.

Fryguy:
@carbonin Any more on this PR?

Fryguy:
@carbonin Bump?

Fryguy:
@carbonin Is this darga yes/no?

carbonin:
@Fryguy it's a no, it definitely isn't a bug.

jrafanie:
🎉

PR #112, opened 2016-03-21
Install netcat and telnet tools

carbonin:
https://bugzilla.redhat.com/show_bug.cgi?id=1318290

### DIFF
--- a/kickstarts/partials/packages/includes.ks.erb
+++ b/kickstarts/partials/packages/includes.ks.erb
@@ -27,6 +27,7 @@ libxslt-devel                    # For nokogiri gem
 lshw                             # From epel
 memcached
 mod_ssl
+nc
 net-snmp
 net-snmp-libs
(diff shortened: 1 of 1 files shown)


carbonin:
@abellotti @blomquisg

abellotti:
I think this is fine, one thing to be aware is that even though telnet would be used for testing/debugging, it does send credentials unencrypted over the wire. @jrafanie thoughts ?

jrafanie:
> I think this is fine, one thing to be aware is that even though telnet would be used for testing/debugging, it does send credentials unencrypted over the wire. @jrafanie thoughts ?

I think this is fine for testing/debugging.  As long as it doesn't start a new insecure daemon that listens for connections, it should be fine to have a new client tool for this purpose.  If we would have to open ports on the firewall to allow traffic, I think that would be a different issue.

PR #113, opened 2016-03-21
Direct output from post to both stdout and tty3

carbonin:
This will allow us to see the output from commands run in the `%post` section in `anaconda-post.log`.

Previously they were being sent to `/dev/tty3` and not stdout so they would not be seen by the `--log` directive.

@abellotti This is the only thing I could get to work outside of removing the `/dev/tty3` stuff entirely.
It's not pretty, but it will be much easier to diagnose build failures this way.

@JPrause FYI

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -38,8 +38,10 @@ reboot
 %end
 
 %post --log=/root/anaconda-post.log
+(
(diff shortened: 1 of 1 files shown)


abellotti:
:+1: @carbonin  may not be pretty, but it gives us details in both vnc into kickstart build as well as log.  will merge when green.

PR #115, opened 2016-03-22
Use full package name for nmap-ncat

carbonin:
This also is the same name as the RHEL package

@jrafanie


jrafanie:
Why?

carbonin:
This is the name of the package that is more commonly used.

PR #119, opened 2016-04-07
Add package hyperv-daemons for vhd appliance

JPrause:
This is to resolve BZ 1322611

@simaishi can you please review.


simaishi:
LGTM

PR #121, opened 2016-04-14
Move compile assets, sti loader and SSUI from cfme-setup

simaishi:
Related PR: https://github.com/ManageIQ/manageiq-appliance/pull/67

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -64,6 +64,8 @@ export RAILS_USE_MEMORY_STORE="true"
 
 <%= render_partial "post/bundler" %>
 
+<%= render_partial "post/ui_compile" %>
+
 # appliance_root="/opt/manageiq/manageiq-appliance" -- in post/source_setup partial
 $appliance_root/cfme-setup.sh
 
--- a/kickstarts/partials/post/ui_compile.ks.erb
+++ b/kickstarts/partials/post/ui_compile.ks.erb
@@ -1,0 +1,15 @@
+npm install gulp bower -g
+
+pushd /var/www/miq/vmdb
+  bower -F --allow-root install
+
+  RAILS_ENV=production rake evm:compile_assets
+  rake evm:compile_sti_loader
+popd
+
+# Self Service UI
+pushd /opt/manageiq/manageiq-ui-self_service
+  npm install
+  bower -F --allow-root install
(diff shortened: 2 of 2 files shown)


kbrock:
@simaishi safe to assume that you have run this on a build? - ok to merge? (aka LGTM)

simaishi:
Yes, ran an appliance build successfully with this + manageiq-appliance change.

kbrock:
On second glance, I think most of this does not belong in kickstart. Thought I realize that install the OS vs the app may be too nebulous.

build|dev|reasoning|example
---|---|---|---
[kickstart]|[guides]|setup os|Install OS, ruby, node, postgres (calls get app code)
[rpm]|[fetch]|download code|Get the app code, put down app files
[cfme-setup.sh]|[setup]|setup app|`bundle update`, `bower install`, `db:create db:migrate seed`
?|[update]|update app|`bundle install`, `db:migrate db:seed`

[rpm]: http://pkgs.devel.redhat.com/cgit/rpms/cfme/tree/cfme.spec?h=cfme-rh-ruby22-5.6-rhel-7
[kickstart]: https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/base.ks.erb
[cfme-setup.sh]: https://github.com/ManageIQ/manageiq-appliance/blob/master/cfme-setup.sh
[fetch]:https://github.com/ManageIQ/guides/blob/master/developer_setup.md#clone-the-code
[guides]: https://github.com/ManageIQ/guides/blob/master/developer_setup.md
[setup]: https://github.com/ManageIQ/manageiq/blob/master/bin/setup
[update]: https://github.com/ManageIQ/manageiq/blob/master/bin/update

@Fryguy does that table look right to you?

Fryguy:
cfme-setup.sh is only run in the kickstart at the moment so moving it directly into kickstart seems fine to me. @kbrock Your table there seems to separate kickstart from rpm from cfme-setup by concern, but in reality they are *all* run in the kickstart.

kbrock:
@Fryguy agreed, merging setup into kickstart probably makes sense.

But the user can still run `rpm update` - so we need to be able to handle that very common use case

Fryguy:
downstream rpm updates are fine as we don't allow bower and bundler updates there (they are provided as part of the various RPMs).

PR #122, opened 2016-04-21
ui_compile - install (newer) npm via npm

himdel:
Because the version of npm available via RPM is only 1.3.6,
and some of the self-service dependencies need npm 3.*,
we'll use the RPM-installed npm to install a current npm version.

Should address https://github.com/ManageIQ/manageiq/issues/8023 and https://github.com/ManageIQ/manageiq/issues/7801

### DIFF
--- a/kickstarts/partials/post/ui_compile.ks.erb
+++ b/kickstarts/partials/post/ui_compile.ks.erb
@@ -1,4 +1,4 @@
-npm install gulp bower -g
+npm install gulp bower npm -g
 
 pushd /var/www/miq/vmdb
   bower -F --allow-root install


himdel:
(Suggesting `darga/yes`, if we do merge..)

jrafanie:
I don't know, @Fryguy @simaishi what do you think?  I'm unsure.

jrafanie:
I'm ok with circumventing rpm packaging temporarily but honestly, but if we don't push epel/software collections to be current in terms of programmer tooling, they'll never be good for developers.

simaishi:
I agree with @jrafanie We can probably put this change in and get the upstream build going, and work on getting the rpm package updated (which I'm sure will take a while).

jrafanie:
Thanks @himdel @simaishi.  Marked for darga backport.

PR #125, opened 2016-04-27
Change upload directory to allow nightly from different branches

simaishi:
Added an option to specify the image upload directory, to allow nightly/release builds from different branches.

This PR will change where the images will be uploaded to and how symlinks are created.  This is aligned with the internal file share structure change @bdunne is working on.

Currently:
- **upstream** dir for any **nightly** build, **latest** symlink for the latest nightly build
- **upstream_stable** dir for any **release** build, **latest** symlink for the latest release build

With this change:
- **master** dir for **nightly** build for **master** branch, **latest** symlink for the latest nightly build
- **\<branch>** dir for **nightly** and **release** builds for **\<branch>**, **latest** symlink for the latest nightly build and **stable** symlink for the latest **release** build
- `--copy-dir <directory name>` to specify upload directory for any build

Once this is merged, fileshare mount on kegerator needs to be updated.

PR #126, opened 2016-04-28
Darga nightly build

simaishi:
Requires #125 to be merged and backported to darga.


abellotti:
I'd expect the script itself to be upstream/master and not a release specific.  Maybe drive the "darga" release/naming based on a settings.yml of some sort so the script does not change with every release.  Thanks.

simaishi:
I thought about that, but the setting has to live somewhere, the script or settings.yml... So I thought there will be no need to introduce another file if we have to change that new file for every release anyway..

I'll change if moving out of the script would be better regardless, though.

PR #128, opened 2016-05-03
Add azure image build

simaishi:
The remaining item is to change disk_layout.

Currently, applinaces have ~6GB swap space on the 40GB disk (--size=4096 --grow).  When a VM is created in Azure, Azure gives an additional disk that can be used to store temporary data at no extra cost (the primary use of this disk is cache/swap). So for Azure, that's where swap will be created:

https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/azure.ks.erb#L3

(set swap to be 10GB for now)

We have a few options for that ~6GB not needed for Azure:

1. Reduce the overall disk size by ~6GB
1. Keep the disk size as is, and increase size on different volume(s) (which one?)
1. Leave everything as is, which makes the swap to be 16GB
1. Don't create swap on temporary storage and make it consistent with other appliances

My preference is 1 or 2, but looking for suggestion... @carbonin @abellotti ?

### DIFF
--- a/config/base_azure.tdl
+++ b/config/base_azure.tdl
@@ -1,0 +1,16 @@
+<template>
+  <name>CentOS-7.2</name>
+  <os>
+    <rootpw>ozrootpw</rootpw>
+    <name>CentOS-7</name>
+    <version>2</version>
+    <arch>x86_64</arch>
+    <install type="iso">
+      <iso>file:///build/isos/CentOS-7-x86_64-DVD-1511.iso</iso>
+    </install>
+  </os>
+  <description>CentOS72 TDL meant to be used with custom kickstart for manageiq.</description>
+  <disk>
+    <size>35G</size>
+  </disk>
+</template>
--- a/kickstarts/partials/main/disk_layout.ks.erb
+++ b/kickstarts/partials/main/disk_layout.ks.erb
@@ -2,15 +2,17 @@
 zerombr
 clearpart       --all --drives=vda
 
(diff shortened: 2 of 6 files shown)


carbonin:
I like option 1, so that would just be a matter of making the size here https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L5 and https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L16 conditional on target type?

simaishi:
plus the overall disk size change here: https://github.com/ManageIQ/manageiq-appliance-build/blob/master/config/base.tdl#L14

That will need to be changed to provide different size based on type

carbonin:
Probably better to just create a new file and use that if the target is azure [here](https://github.com/ManageIQ/manageiq-appliance-build/blob/master/scripts/vmbuild.rb#L73)

carbonin:
Unless we want to make that `base.tdl.erb` and evaluate that like we do with the kickstarts ...

simaishi:
Based on suggestion by @carbonin and discussion with @abellotti, the disk size is reduced for Azure and Azure will use different base.tdl.  If we're going to customize base.tdl more in the future, we'll go for erb route.

Since the swap was 5.86GB, the disk size was reduced by 5GB.

abellotti:
👍 looks good.

PR #129, opened 2016-05-04
Use ruby 2.2.5 for appliance builds.

jrafanie:
cc @bdunne @Fryguy

I'm thinking darga should go out with the latest 2.2.x ruby.


jrafanie:
cc @simaishi

bdunne:
:+1: LGTM

PR #132, opened 2016-05-13
Use build type to determine nightly or release

simaishi:
Now we have 'darga' nightly, we can't rely on the word 'master' being in the file name to determine if it's a nightly build or not.

@abellotti @Fryguy please review


abellotti:
LGTM!!  👍

PR #133, opened 2016-05-18
docbuild.rb needs to create latest-docs/master/ in target for ascii_binder.

abellotti:
ascii_binder references the stylesheet via ../../master/ so we need the latest-docs to include the master/ subdirectory.

### DIFF
--- a/scripts/docbuild.rb
+++ b/scripts/docbuild.rb
@@ -30,5 +30,6 @@ $log.info("Packaging docs...Complete")
 
 $log.info("Copying docs to fileshare...")
 FileUtils.rm_rf DOCS_FILESHARE
+FileUtils.mkdir_p DOCS_FILESHARE
 FileUtils.cp_r(DOCS_PACKAGE, DOCS_FILESHARE)
 $log.info("Copying docs to fileshare...Complete")


abellotti:
@simaishi @Fryguy minor update to docbuild.rb to keep ascii_binder happy. Thanks.

scripts/docbuild.rb
```diff
@@ -30,5 +30,6 @@ def execute(cmd)
 
 $log.info("Copying docs to fileshare...")
 FileUtils.rm_rf DOCS_FILESHARE
+FileUtils.mkdir_p DOCS_FILESHARE
```
Fryguy: I am intentionally copying the DOCS_PACKAGE to DOCS_FILESHARE to remove the level of indirection, though maybe that's me being overly cautious...I think this is actually a bug in asciibinder when it packages.

abellotti: wasn't sure if it was an asciibinder bug, all works well if you browse the built stuff under _package/community/, it has the master/ subdirectory there.

abellotti: @Fryguy any update on this ? Can we merge this for now at least until there's a fix in asciibinder land ? (maybe nhr can help there). Thanks.

PR #137, opened 2016-06-24
Use the centos.org repo for scl packages

carbonin:
This repo seems to be updated more frequently than the copr repo and contains more recent packages.

This will also allow us to pull PostgreSQL 9.5, which the other repo did not have.

### DIFF
--- a/kickstarts/partials/main/repos.ks.erb
+++ b/kickstarts/partials/main/repos.ks.erb
@@ -5,7 +5,7 @@ repo --name=extras  --baseurl=http://mirror.centos.org/centos/7/extras/x86_64/
 repo --name=epel    --baseurl=http://dl.fedoraproject.org/pub/epel/7/x86_64/
 
 # repos to install packages not found in the os
-repo --name=rh-postgresql94_copr --baseurl=https://copr-be.cloud.fedoraproject.org/results/rhscl/rh-postgresql94/epel-7-x86_64/
+repo --name=centos-scl --baseurl=http://mirror.centos.org/centos/7/sclo/x86_64/rh/
 repo --name=pglogical-scl --baseurl=https://copr-be.cloud.fedoraproject.org/results/ncarboni/pglogical-SCL/epel-7-x86_64/
 <% if @target == "gce" %>
 repo --name=google-cloud-compute --baseurl=https://packages.cloud.google.com/yum/repos/google-cloud-compute-el7-x86_64/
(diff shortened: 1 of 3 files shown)


carbonin:
After adding this repo, we are pulling additional nodejs packages that we were not getting before.  Specifically `nodejs010-libuv.x86_64` and `nodejs010-runtime.x86_64`

These show up as a dependency of `nodejs` only after adding the repo.
Before:
```
[root@localhost ~]# repoquery --requires --resolve nodejs | sort
bash-0:4.2.46-19.el7.x86_64
ca-certificates-0:2015.2.6-70.1.el7_2.noarch
glibc-0:2.17-105.el7.i686
glibc-0:2.17-106.el7_2.6.x86_64
libgcc-0:4.8.5-4.el7.x86_64
libstdc++-0:4.8.5-4.el7.x86_64
libuv-1:0.10.34-2.el7.x86_64
openssl-libs-1:1.0.1e-51.el7_2.4.x86_64
zlib-0:1.2.7-15.el7.x86_64
```

After:
```
[root@localhost anaconda]# repoquery --requires --resolve nodejs | sort
bash-0:4.2.46-19.el7.x86_64
ca-certificates-0:2015.2.6-70.1.el7_2.noarch
glibc-0:2.17-105.el7.i686
glibc-0:2.17-106.el7_2.6.x86_64
libgcc-0:4.8.5-4.el7.x86_64
libstdc++-0:4.8.5-4.el7.x86_64
libuv-1:0.10.34-2.el7.x86_64
nodejs010-libuv-1:0.10.30-1.el7.x86_64
openssl-libs-1:1.0.1e-51.el7_2.4.x86_64
zlib-0:1.2.7-15.el7.x86_64
```

This addition seems to be the cause of some messages being printed when a login shell is started:
`Can't read /etc/scl/conf/v8314, v8314 is probably not installed.`

I'm not familiar enough with the significance of these packages to know if this is indicative on a larger issue or if it can be safely ignored

@jrafanie @simaishi ?

carbonin:
@miq-bot add_label wip, core

Fryguy:
Wha?  So epel's nodejs (non) has nodejs010 (scl) as a depedency?! This is weird @bdunne @simaishi

carbonin:
This might be being caused by installing the repo file in the packages section rather than adding it in the post like the other repos here https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/repos.ks.erb

carbonin:
Looks like the nodejs010 packages being installed won't be a problem. I will just remove the call to enable [here](https://github.com/ManageIQ/manageiq-appliance/blob/02e2439a0fb2bfc907bbc97e3500e858e70c281d/LINK/etc/default/evm#L22) in a follow up which will remove the error message at login.

carbonin:
The errors on login will be fixed when https://github.com/ManageIQ/manageiq-appliance/pull/75 is merged

carbonin:
@miq-bot add_label enhancement, darga/no

gtanzillo:
👍 LGTM

PR #138, opened 2016-06-24
Clean the rails log directory in the kickstart post

carbonin:
Some things that are done during the build process have the potential to print to the logs.
None of this should be present when the machine is deployed for the first time.

https://bugzilla.redhat.com/show_bug.cgi?id=1286793

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -99,6 +99,9 @@ ramfsfile="/boot/initramfs-$kversion.img"
 # make sure there is a new line at the end of sshd_config
 echo "" >> /etc/ssh/sshd_config
 
+# Clean the logs
+rm -rf "$vmdb_root/log/*"
+
 chvt 1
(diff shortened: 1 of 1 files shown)


carbonin:
@miq-bot add_label darga/yes, bug

gtanzillo:
👍 LGTM

kickstarts/base.ks.erb
```diff
@@ -99,6 +99,9 @@ ramfsfile="/boot/initramfs-$kversion.img"
 # make sure there is a new line at the end of sshd_config
 echo "" >> /etc/ssh/sshd_config
 
+# Clean the logs
+rm -rf "$vmdb_root/log/*"
```
Fryguy: I just noticed this, and I think we should use the trick that Geert used where we mount tmpfs, and then just toss the tmpfs disk at the end.  This way we keep the image size down since deletes don't really delete the on-disk bits. cc @simaishi

PR #139, opened 2016-07-06
Explicitly include lvm2 package

carbonin:
We require lvm2 to be installed when initializing the database so it should be explicitly included in the kickstart.

https://bugzilla.redhat.com/show_bug.cgi?id=1343007


carbonin:
@gtanzillo I think this one is good to go.

gtanzillo:
LGTM 👍

PR #143, opened 2016-08-04
Add libcurl-devel, needed by curb

simaishi:
`libcurl-devel` is needed by `curb` gem which is a dependency of `ovirt-engine-sdk` added in https://github.com/ManageIQ/manageiq/pull/9521

Fixes https://github.com/ManageIQ/manageiq/issues/10234


carbonin:
:+1: Looks good. Just hit this myself. @gtanzillo

gtanzillo:
👍

PR #144, opened 2016-08-08
Add support for building Vagrant appliances.

geertj:
This patch series adds support for building Vagrant "boxes". There's two patches in the series, one prepares support for multiple OVA formats, and the second patch uses that to add in Vagrant support.

Vagrant boxes are (Vagrant) provider specific. This builds a box for the VirtualBox provider, which is the most common one. Other formats can be added later. Also no upload to Atlas is implemented yet.

### DIFF
--- a/config/target.json
+++ b/config/target.json
@@ -1,5 +1,3 @@
 {
-  "generate_icicle": false,
-  "openstack_image_format": "qcow2",
-  "rhevm_image_format": "qcow2"
+  "generate_icicle": false
 }
--- a/config/target.json
+++ b/config/target.json
@@ -1,0 +1,5 @@
+{
+  "generate_icicle": false,
+  "openstack_image_format": "qcow2",
+  "rhevm_image_format": "qcow2"
+}
--- a/kickstarts/partials/post/vagrant.ks.erb
+++ b/kickstarts/partials/post/vagrant.ks.erb
@@ -1,0 +1,23 @@
+# Add a user "vagrant" with password "vagrant" and a well known public key.
+# Note this is how Vagrant works. You are assumed not to use a Vagrant box on
+# a publicly acessible network.
+useradd vagrant -p b1HEow3cJo7Nc
+
+cat << EOM > /etc/sudoers.d/vagrant-nopasswd
+Defaults:vagrant !requiretty
+vagrant ALL=(ALL) NOPASSWD:ALL
+EOM
+sed -i 's/.*UseDNS.*/UseDNS no/' /etc/ssh/sshd_config
+chmod 400 /etc/sudoers.d/vagrant-nopasswd
+
+mkdir -m 0700 -p ~vagrant/.ssh
+cat << EOM > ~vagrant/.ssh/authorized_keys
+ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ== vagrant insecure public key
+EOM
+chmod 600 ~vagrant/.ssh/authorized_keys
+chown -R vagrant:vagrant ~vagrant/.ssh/
+
+cat << EOM >> ~vagrant/.bashrc
+umask 022
+alias appliance_console='sudo appliance_console'
+EOM
--- a/scripts/kickstart_generator.rb
+++ b/scripts/kickstart_generator.rb
@@ -23,17 +23,17 @@ module Build
       @ssui_checkout      = ssui_checkout
     end
 
-    def run(task = :all)
+    def run
+      FileUtils.mkdir_p(@ks_gen_base)
+
       targets.each do |target|
         @target = target # used during ERB evaluation
 
         result = evaluate_erb
 
-        FileUtils.mkdir_p(@ks_gen_base)
-        File.write(@ks_gen_base.join("base.ks"), result)
-
-        write_config(result) if [:all, :config].include?(task)
-        write_json(result)   if [:all, :json].include?(task)
+        file = @ks_gen_base.join("base-#{@target}.ks")
+        $log.info("Writing kickstart file: #{file}") if $log
+        File.write(file, result)
       end
     end
 
@@ -42,23 +42,6 @@ module Build
     end
 
     private
-
-    def write_config(result)
-      file = @ks_gen_base.join("base-#{@target}.cfg")
-      $log.info("Writing kickstart in config format: #{file}") if $log
-      File.write(file, result)
-    end
-
-    def write_json(result)
-      json = {
-        "install_script"  => result,
-        "generate_icicle" => false
-      }.to_json
-
-      file = @ks_gen_base.join("base-#{@target}.json")
-      $log.info("Writing kickstart in json format: #{file}") if $log
-      File.write(file, json)
-    end
 
     def evaluate_erb
       ks_file = Productization.file_for(@build_base, "#{KS_DIR}/base.ks.erb")
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -70,6 +70,8 @@ end
 
 $log.info "Using Configuration base directory: #{cfg_base}"
 
+base_file = BUILD_BASE.join("config/base.json")
+target_file = BUILD_BASE.join("config/target.json")
 ova_file = BUILD_BASE.join("config/ova.json")
 
 def verify_run(output)
@@ -109,48 +111,43 @@ FileUtils.mkdir_p(destination_directory)
 Dir.chdir(IMGFAC_DIR) do
   targets.sort.reverse.each do |target|
     imgfac_target = target.imagefactory_type
-    vhd_image     = target.file_extension == "vhd"
+    ova_format    = target.ova_format
     $log.info "Building for #{target}:"
 
-    tdl_name = imgfac_target == "azure" ? "base_azure.tdl" : "base.tdl"
+    tdl_name = target.name == "azure" ? "base_azure.tdl" : "base.tdl"
     tdl_file = BUILD_BASE.join("config", tdl_name)
     $log.info "Using inputs: puddle: #{puddle}, build_label: #{build_label}"
     $log.info "              tdl_file: #{tdl_file}, ova_file: #{ova_file}."
 
-    input_file  = ks_gen.gen_file_path("base-#{target}.json")
-    output_file = ks_gen.gen_file_path("base-#{target}-#{build_label}-#{timestamp}.json")
+    input_file  = ks_gen.gen_file_path("base-#{target}.ks")
+    output_file = ks_gen.gen_file_path("base-#{target}-#{build_label}-#{timestamp}.ks")
 
     FileUtils.cp(input_file, output_file)
 
-    log_params = "kickstart: #{output_file} copied from #{input_file}. tdl: #{tdl_file}"
-    $log.info "Running base_image using parameters: #{log_params}"
+    params = "--parameters #{base_file} --file-parameter install_script #{output_file}"
+    $log.info "Running #{target} base_image using parameters: #{params}"
 
-    output = `./imagefactory --config #{IMGFAC_CONF} base_image --parameters #{output_file} #{tdl_file}`
+    output = `./imagefactory --config #{IMGFAC_CONF} base_image #{params} #{tdl_file}`
     uuid   = verify_run(output)
     $log.info "#{target} base_image complete, uuid: #{uuid}"
 
-    unless vhd_image
-      $log.info "Running #{target} target_image with #{imgfac_target} and uuid: #{uuid}"
-      output = `./imagefactory --config #{IMGFAC_CONF} target_image --id #{uuid} #{imgfac_target}`
-      uuid   = verify_run(output)
-      $log.info "#{target} target_image with imgfac_target: #{imgfac_target} and uuid #{uuid} complete"
+    params = "--parameters #{target_file}"
+    $log.info "Running #{target} target_image #{imgfac_target} using parameters: #{params}"
 
-      unless imgfac_target == "openstack-kvm"
-        $log.info "Running #{target} target_image ova with ova file: #{ova_file} and uuid: #{uuid}"
-        output = `./imagefactory --config #{IMGFAC_CONF} target_image ova --parameters #{ova_file} --id #{uuid}`
-        uuid   = verify_run(output)
-        $log.info "#{target} target_image ova with uuid: #{uuid} complete"
-      end
+    output = `./imagefactory --config #{IMGFAC_CONF} target_image #{params} --id #{uuid} #{imgfac_target}`
+    uuid   = verify_run(output)
+    $log.info "#{target} target_image #{imgfac_target} complete, uuid: #{uuid}"
+
+    if ova_format
+      params = "--parameters #{ova_file} --parameter #{imgfac_target}_ova_format #{ova_format}"
+      $log.info "Running #{target} target_image ova using parameters: #{params}"
+
+      output = `./imagefactory --config #{IMGFAC_CONF} target_image #{params} --id #{uuid} ova`
+      uuid   = verify_run(output)
+      $log.info "#{target} target_image ova complete, uuid: #{uuid}"
     end
     $log.info "Built #{target} with final UUID: #{uuid}"
     source = STORAGE_DIR.join("#{uuid}.body")
-
-    if vhd_image
-      $log.info "Running qemu-img to convert the raw image"
-      source_converted = STORAGE_DIR.join("#{uuid}.converted")
-      $log.info `qemu-img convert -f raw -O vpc #{source} #{source_converted}`
-      source = source_converted
-    end
 
     FileUtils.mkdir_p(destination_directory)
     file_name = "#{name}-#{target}-#{build_label}-#{timestamp}-#{manageiq_checkout.commit_sha}.#{target.file_extension}"
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 /pkg
 /productization
-kickstarts/generated/base.ks
-kickstarts/generated/*.json
-kickstarts/generated/*.cfg
+kickstarts/generated/*.ks
 scripts/Gemfile.dev.rb
--- a/README.md
+++ b/README.md
@@ -97,7 +97,8 @@ Below are instructions on installing and configuring a virtual machine to genera
 
 ## Setup Imagefactory:
 
-  * Clone imagefactory as /build/imagefactory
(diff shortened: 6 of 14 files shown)


scripts/kickstart_generator.rb
```diff
@@ -23,17 +23,17 @@ def initialize(build_base, targets, puddle, manageiq_checkout, appliance_checkou
       @ssui_checkout      = ssui_checkout
     end
 
-    def run(task = :all)
+    def run
```
gtanzillo: This change seems to be breaking a bunch of tests that expect run to receive `task` as an arg.

scripts/kickstart_generator.rb
```diff
@@ -43,23 +43,6 @@ def gen_file_path(file)
 
     private
 
-    def write_config(result)
-      file = @ks_gen_base.join("base-#{@target}.cfg")
-      $log.info("Writing kickstart in config format: #{file}") if $log
-      File.write(file, result)
-    end
-
-    def write_json(result)
```
gtanzillo: Is this no longer necessary?

scripts/kickstart_generator.rb
```diff
@@ -23,17 +23,17 @@ def initialize(build_base, targets, puddle, manageiq_checkout, appliance_checkou
...
         result = evaluate_erb
 
-        FileUtils.mkdir_p(@ks_gen_base)
-        File.write(@ks_gen_base.join("base.ks"), result)
-
-        write_config(result) if [:all, :config].include?(task)
-        write_json(result)   if [:all, :json].include?(task)
+        file = @ks_gen_base.join("base-#{@target}.ks")
+        $log.info("Writing kickstart file: #{file}") if $log
+        File.write(file, result)
```
jrafanie: We needed config and json format at one point, while it's worthwhile checking if we still need it, we shouldn't change this as part of this feature.  

Note, I believe imagefactory/koji at one point required one type of file as input and then needed a different format.  If we only ever use one format for both imagefactory/koji builds, I'm good with removing the "unused" format in a separate PR.  CC @simaishi @carbonin

scripts/kickstart_generator.rb
jrafanie: See above, we should really try to get this feature working with the existing json/config formats and determine if we still need both of those formats in a future PR.

kickstarts/partials/post/vagrant.ks.erb
```diff
@@ -0,0 +1,23 @@
+# Add a user "vagrant" with password "vagrant" and a well known public key.
+# Note this is how Vagrant works. You are assumed not to use a Vagrant box on
+# a publicly acessible network.
```
jrafanie: This feels like a big assumption.  I'm not familiar with vagrant to understand why this is needed.

scripts/vmbuild.rb
```diff
@@ -109,49 +111,44 @@ def verify_run(output)
...
+      uuid   = verify_run(output)
+      $log.info "#{target} target_image ova complete, uuid: #{uuid}"
     end
     $log.info "Built #{target} with final UUID: #{uuid}"
     source = STORAGE_DIR.join("#{uuid}.body")
 
-    if vhd_image
-      $log.info "Running qemu-img to convert the raw image"
-      source_converted = STORAGE_DIR.join("#{uuid}.converted")
-      $log.info `qemu-img convert -f raw -O vpc #{source} #{source_converted}`
```
jrafanie: Where is this step in the new code?

scripts/vmbuild.rb
```diff
@@ -109,49 +111,44 @@ def verify_run(output)
...
-      end
+    output = `./imagefactory --config #{IMGFAC_CONF} target_image #{params} --id #{uuid} #{imgfac_target}`
+    uuid   = verify_run(output)
+    $log.info "#{target} target_image #{imgfac_target} complete, uuid: #{uuid}"
+
+    if ova_format
+      params = "--parameters #{ova_file} --parameter #{imgfac_target}_ova_format #{ova_format}"
+      $log.info "Running #{target} target_image ova using parameters: #{params}"
+
+      output = `./imagefactory --config #{IMGFAC_CONF} target_image #{params} --id #{uuid} ova`
```
jrafanie: Do any of these changes/added CLI options depend on having a newer version of imagefactory?

jrafanie:
@simaishi @JPrause Please review, especially the changes around image formats and options we pass to imagefactory.  Thanks!

scripts/kickstart_generator.rb
geertj: Previously imagefactory only had a single `--parameters` option, which read a single JSON file containing options. This means that if you had multiple options, you had to stuff them all in there. The result was that the kickstart generator had to write the kickstart file as well as non-kickstart related options in there.

Newer versions have added multiple `--parameter` and `--file-parameter` options (still only a single `--parameter*s*` option though). By using this we can implement a separation of concerns where the kickstart generator only outputs the kickstart file, and we can pass in other options via additional command line arguments.

I could do this in separate PRs but given that I'm already restructuring  part of `vmbuild.rb` I'd prefer not to have to do that extra work. Note that the Vagrant support is a separate commit in the same PR. So it's already somewhat separated.

I've tested the code and all formats build correctly. So I'm pretty sure that we don't need the "Kickstart in JSON" workaround anymore.

scripts/kickstart_generator.rb
geertj: See my comments above. I've restructured `vmbuild.rb` in a separate commit in this PR. I could leave this code it but it would be unused code.

scripts/vmbuild.rb
geertj: Yes, you need the latest upstream version of imagefactory. Also note that I ran into a bug in Oz (unrelated to my changes). I cannot get Oz to work reliably without this PR:

https://github.com/clalancette/oz/pull/217

scripts/vmbuild.rb
geertj: See my comment at the top of this commit (I want to make sure you're reviewing this as two commits in a single PR, instead of one big commit. The Github interface makes this somewhat unintuitive.). Latest imagefactory supports VHD format and so we can remove this hack and do a clean 3-phase build process where all phases use imagefactory.

scripts/kickstart_generator.rb
geertj: I didn't realize that there was a test harness for this code.  I'll have a look at fixing it.

kickstarts/partials/post/vagrant.ks.erb
geertj: See: https://www.vagrantup.com/docs/boxes/base.html Vagrant needs to be able to log in to virtual machines, and for that it requires the insecure key pair, and default password. All vagrant boxes work like this.

Note that Vagrant *does* create a box behind a  NAT with no inbound access by default.  You have to do real work in order to expose it publicly.

scripts/kickstart_generator.rb
geertj: Tests have been updated.

scripts/vmbuild.rb
Fryguy: So this PR would require us to bump imagefactory on the build machine?  cc @simaishi

simaishi: This will definitely require us to update imagefactory on the build machine.  We should probably consider using a tag and specif exactly which version of imagefactory is required to avoid confusion.

config/target.json
```diff
@@ -0,0 +1,5 @@
+{
+  "generate_icicle": false,
+  "openstack_image_format": "qcow2",
+  "rhevm_image_format": "qcow2"
```
simaishi: With the image format settings in this file now, can we remove them from imagefactory.conf?
https://github.com/ManageIQ/manageiq-appliance-build/blob/master/config/imagefactory.conf#L8-L9

scripts/vmbuild.rb
Fryguy: @geertj The version needs to be added to the setup in the README as well.

config/target.json
jrafanie: That would be nice but I don't know.

kickstarts/partials/post/vagrant.ks.erb
jrafanie: ok, thanks

scripts/kickstart_generator.rb
jrafanie: Ok, thanks @geertj.  I like the "new" way, having two formats was silly.  I don't know that we need both formats still though or if we can assume brew/koji and our upstream build machine vms have the required imagefactory versions to take advantage of this.

jrafanie: cc @simaishi

scripts/vmbuild.rb
jrafanie: @simaishi @chrispy1 @JPrause do we use any of this code in brew/koji because I don't know that they'll have this new version of imagefactory yet?  Also, we'll have to update our build vms to use this.

scripts/vmbuild.rb
jrafanie: Ok, cool @simaishi do you know how to verify the raw conversion process once we setup the new imagefactory?  I don't remember why it was added.

simaishi: @jrafanie Since the actual qemu-img command used is same before/after the change, I'd expect this to work without problem, but I'll verify the converted vhd.

scripts/vmbuild.rb
simaishi: This part of the code is upstream only and won't affect brew build, so we're ok there.

scripts/vmbuild.rb
geertj: @Fryguy correct, we need the latest imagefactory, will add this to the README.

config/target.json
geertj: Yup, we can remove it from there. I will update the PR to include that change.

scripts/vmbuild.rb
geertj: @Fryguy actually the README already states to use the upstream master, see: https://github.com/ManageIQ/manageiq-appliance-build/blob/master/README.md#setup-imagefactory

simaishi: @geertj  Can we use a tag instead? It would be a lot easier to identify which PR requires which imagefactory code when using a tag as we can update README at the same time.

scripts/vmbuild.rb
jrafanie: @simaishi Are you saying that none of the code changes in this PR affect the downstream build because of the imagefactory version change?

scripts/vmbuild.rb
jrafanie: Yeah, I agree.  It would be good to get them to release a tag and push to get koji/brew to use this same tag at some point.

scripts/kickstart_generator.rb
jrafanie: I'm ok with doing this as long as works for all of our upstream images and doesn't affect downstream builds.  I'll wait for @simaishi's approval.

scripts/vmbuild.rb
simaishi: The only thing in this PR that will affect the downstream build is the kickstart_generator.rb change.  I can take care of that once this gets merged.

The changes in vmbuild.rb is all done in brew for downstream and it won't affect our downstream code.

jrafanie: Thanks @simaishi!  So, then, my only reservation is we need a imagefactory git sha (prefer an actual tag) and we'll need to get the build vm(s) updated to use this imagefactory.  I don't know how much work this since I'm not sure if any dependencies of imagefactory have changed.

geertj: I have added a mention of the minimum git sha to the README. (I have no permissions to create tags in imagefactory.)

jrafanie: @simaishi @chrispy1 I'm not sure what you want do but I would feel better having the oz PR merged, get a release tagged, possibly packaged, and get a tag for imagefactory.  Our existing instructions use imagefactory from source, so using a specific commit sha should be doable although a tag would be better.  But, we install oz from rpm with yum so we've not run oz on our build machines from source.  I'm not sure what you want to do.

These are the [instructions](https://github.com/ManageIQ/manageiq-appliance-build/blob/1ea4ac11b2c8e70494d748ffde434f8c988e3c97/README.md#setup-imagefactory) we wrote many months ago that would need to be updated if we change how we set it up.

simaishi: @jrafanie That's what I was thinking too.. Don't think it's a good idea to use a non-merged change, and I agree with you regarding using code from source vs tag.

I updated my local build machine to use the latest imagefactory.  Without the oz change, it's building fine so far.  But my build machine is Fedora, oz is installed from rpm and I didn't update oz.  Not sure which version of oz might cause that issue that requires a fix in oz.

geertj: @jrafanie @simaishi - OK, trying to summarize and hopefully close the remaining loose ends:

1. My PR to Oz is not merged. I pinged @clalancette on clalancette/oz#217. HOWEVER - the issue is not related to this PR. It is a race condition that on my system is triggered for both the current builder, and well as the builder with my changes. Maybe the older, RPM version packaged in EPEL/Fedora is not impacted, I don't know. But I always hit the issue with the Oz master, irrespective of my change.

2. Requirement for the need for a tag of imagefactory.  I had a look and there are no new tags to it since 2014. We can ask @imcleod to create a new version/tag. However since the current instructions in the builder README already require you to use the latest upstream, there is no change and I if this is needed it can be done separately.

In summary,  none of these issues is related to my patch, and it would be great if this could be merged :) I'm going to have some more time this weekend for hacking but what I'm planning (upload to Vagrant Atlas) will depend on this PR.

scripts/vmbuild.rb
imcleod: I stopped regularly tagging after we change the Makefile and RPM process to use the version embedded in setup.py.  I'm happy to go back to tagging/updating on release boundaries.  For now, I've tagged the latest upstream (which also corresponds to the latest Fedora and EPEL releases) with the RPM VR:

https://github.com/redhat-imaging/imagefactory/releases/tag/1.1.9-3

scripts/vmbuild.rb
jrafanie: >In summary, none of these issues is related to my patch, and it would be great if this could be merged :)

@geertj we need to upgrade imagefactory to accept this patch AFAIK which it sounds like we might also need to the oz patch, so I think it does relate to this PR.  If you can make the changes without an imagefactory update, then yes, we can merge it right away.  

If we need to update our build vms, the docs, and anything else to use imagefactory at a specific tag or commit (with perhaps the change for oz too), I'll need @simaishi @JPrause ack to merge it since they know more about the build vms and those environments.

geertj: Thanks @jrafanie! @simaishi @JPrause what is required to update the imagefactory on the build VM. Anything I can do to help? I wouldn't mind helping out with the update or even doing it, if you are open to that.

jrafanie: Thanks @geertj.  I'll wait for Satoe to reply.  

Note, even though we use imagefactory from source, we don't often if ever `git pull` on the build vm.   Usually, we only pull imagefactory if we need a feature. On the other hand, we `git pull` from this repo, `manageiq-appliance-build`, all the time so any changes we merge here will land for this repo but not for imagefactory.

simaishi: I'll be looking at that today.  I'm testing locally first, then update our production build machine (after taking a snapshot!).  Since our build machine builds all branches, we need to make sure both master and darga can be still built properly after the update.

The required SHA is same as 1.1.9, so I'm thinking we should probably use RPM instead of getting from source.  Also I won't be touching oz because of the issue @geertj mentioned (the build machine has 0.14.0, the latest is 0.15.0).  So the update will be only to imagefactory and its dependency if any.  @jrafanie sounds ok?

jrafanie: @simaishi Thanks! That sounds great.  Minimize the potential problems.  Maybe oz will be fixed and released by the time we need to use that version.

abellotti: @simaishi sounds like a good plan. I'm not sure of the 0.15.0 python package dependencies and if we need to update any of those RPMs.  definitely a snapshot first :)  Are you testing on your Fedora or CentOS personal build machine first ?  Preferring the later. You're welcome to use my build machine for testing this if you need it.  Thanks.

geertj:
Squashed the small follow-on patches.

clalancette:
Maintainer of Oz here.  I've been away on holiday for a bit; I *may* have
time this week to review the patch from Geert, but I can't promise that.
If not, it will have to wait another couple of weeks, but I'll definitely
get to it in late August/early September.

Chris Lalancette

scripts/vmbuild.rb
```diff
@@ -109,49 +111,44 @@ def verify_run(output)
 Dir.chdir(IMGFAC_DIR) do
   targets.sort.reverse.each do |target|
     imgfac_target = target.imagefactory_type
-    vhd_image     = target.file_extension == "vhd"
+    ova_format    = target.ova_format
     $log.info "Building for #{target}:"
 
     tdl_name = imgfac_target == "azure" ? "base_azure.tdl" : "base.tdl"
```
simaishi: @garrett This will need to be changed as the value of `imagefactory_type` is changed for azure in this PR.  We could have just checked for `target` anyway..

geertj: @simaishi great catch! Should have caught this myself. Will update the PR tonight.

config/target.json
geertj: Hmm, that is weird. See:

https://github.com/redhat-imaging/imagefactory/blob/master/imagefactory#L180

Here you see all 3 command line options that add/remove configuration being merged into `ApplicationConfig().configuration`. I will do some tests tonight and see what is happening. Did you try it out, or were you reasoning about the code?

geertj:
Thanks to @clalancette the fix to Oz has been merged into the Oz master:

https://github.com/clalancette/oz/commit/ad7d85178e2de9f463133b6a63953012db822657

As discussed we don't appear to need this on our build machine as Oz 0.14.0 appears to work, so just FYI.

config/target.json
simaishi: I tried it out.  For both RHEVM and OpenStack, it didn't go through the conversion unless I put the settings back to imagefactory.conf.

scripts/vmbuild.rb
Fryguy: Any way we can get a test to catch that, or is that too complicated as this is the actual building script?

scripts/vmbuild.rb
geertj: README updated to refer to tag.

config/target.json
geertj: @simaishi fixed. See the comment in the PR about why it didn't work.

scripts/vmbuild.rb
geertj: @simaishi - fixed.

@Fryguy NOT easy to test this :)

geertj:
@simaishi let me know if there's anything else! Also do you want me to squash the fixup commits?

simaishi:
I was going to switch to RPM, but that will require all the plugins to be separately installed as RPMs as well and I thought that's beyond the scope of this PR.  So I just updated imagefactory source to use 1.1.9-3 and tested successfully for master and darga branches.

Our production build machine is now updated to 1.1.9-3 and ready to take new formats used in this PR.

The only thing is.. after updating to 1.1.9-3, running /build/bin/setup_imagefactory.sh didn't create a symlink for HyperV in /etc/imagefactory/plugins.d directory.  The workaround is to add HyperV to the list of plugins in imagefactory/scripts/imagefactory_dev_setup.sh (which is called by  setup_imagefactory.sh).  @geertj We should add this to README, as hyperv/azure builds won't work without this.

setup_imagefactory.sh needs a bit of cleanup.  I'll either look into cleaning that up OR switching to RPM later.

simaishi:
@geertj Thanks for the changes, LGTM!  @jrafanie @abellotti ?

jrafanie:
Thanks @geertj @simaishi !!!  LGTM and since you both tried it out, merging.

jrafanie:
👏 🍪 🎉

Fryguy:
:+1: Great work @geertj !

PR #146, opened 2016-08-22
Add support for libvirt and GCE

geertj:
Add support for building libvirt and GCE appliance.

These are two separate commits but in one PR. The commits step on each other so it's easier to have it in one PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,8 @@ Below are instructions on installing and configuring a virtual machine to genera
 
 ## Setup Imagefactory:
 
-  * Clone imagefactory as /build/imagefactory (requires version 1.1.9-3 or
-    later for command line parameter passing and hyperv target support):
+  * Clone imagefactory as /build/imagefactory (requires git tag f13dc65
+    or later for google compute engine support):
 
     ```
     cd /build
@@ -141,10 +141,11 @@ Below are instructions on installing and configuring a virtual machine to genera
     ```
 
   * Note that there's a bug in imagefactory where it doesn't install the
-    HyperV plugin. Until that is fixed, use the following command to install
-    it.
+    HyperV and GCE plugins. Until that is fixed, use the following commands to
+    install them.
 
     # ln -s /build/imagefactory/imagefactory_plugins/HyperV/HyperV.info /etc/imagefactory/plugins.d
+    # ln -s /build/imagefactory/imagefactory_plugins/GCE/GCE.info /etc/imagefactory/plugins.d
 
 ## Setup for vSphere plugin.
 
--- a/kickstarts/partials/packages/includes.ks.erb
+++ b/kickstarts/partials/packages/includes.ks.erb
@@ -91,6 +91,8 @@ npm
 <% if @target == "ovirt" %>
 ovirt-guest-agent
 glusterfs-fuse
+<% elsif @target == "libvirt" %>
+qemu-guest-agent
 <% elsif @target == "vsphere" %>
 open-vm-tools
 <% elsif @target == "azure" %>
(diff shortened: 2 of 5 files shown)


geertj:
@simaishi can you have a look at this?

kickstarts/partials/packages/includes.ks.erb
```diff
@@ -91,6 +91,8 @@ npm
 <% if @target == "ovirt" %>
 ovirt-guest-agent
 glusterfs-fuse
+<% elsif @target == "libvirt" %>
+qemu-guest-agent
```
Fryguy: Minor but in a followup PR can we sort this and use a case statement? cc @simaishi

scripts/target.rb
```diff
@@ -8,7 +8,9 @@ class Target
       'openstack' => ImagefactoryMetadata.new('openstack-kvm', nil, 'qc2'),
       'hyperv'    => ImagefactoryMetadata.new('hyperv', nil, 'vhd'),
       'azure'     => ImagefactoryMetadata.new('hyperv', nil, 'vhd'),
-      'vagrant'   => ImagefactoryMetadata.new('vsphere', 'vagrant-virtualbox', 'box')
+      'vagrant'   => ImagefactoryMetadata.new('vsphere', 'vagrant-virtualbox', 'box'),
+      'libvirt'   => ImagefactoryMetadata.new('openstack-kvm', nil, 'qc2'),
```
Fryguy: Isn't this the same as the openstack image?  Would it be simpler to just rename the openstack image or copy it instead of a full build?

geertj: @Fryguy the target name drives certain conditional statements in the kickstarts. So while the definition looks identical, it will result in a different appliance.

kickstarts/partials/packages/includes.ks.erb
geertj: Happy to do that. Still earning Ruby :)

scripts/target.rb
```diff
@@ -8,7 +8,9 @@ class Target
...
+      'gce'       => ImagefactoryMetadata.new('gce', nil, 'tgz'),
```
simaishi: @geertj Can you change the extension to`.tar.gz`?  That's GCE's requirement and disk can't be created from `.tgz` image.

README.md
```diff
@@ -97,8 +97,8 @@ Below are instructions on installing and configuring a virtual machine to genera
 
 ## Setup Imagefactory:
 
-  * Clone imagefactory as /build/imagefactory (requires version 1.1.9-3 or
-    later for command line parameter passing and hyperv target support):
+  * Clone imagefactory as /build/imagefactory (requires git tag f13dc65
+    or later for google compute engine support):
```
simaishi: The GCE support also has the same symlink problem, just like HyperV.  Can you please add a note about manually creating symlink for GCE?

geertj:
@simaishi I've added the note about the symlink and changed the file extension to .tar.gz. The reason that I didn't run into the file extension issue is that I used the recently merged capability in imagefactory to push directly to GCE. That code adds a `.tar.gz`  file extension automatically. My next PR will enable this in manageiq-appliance-build.

@Fryguy let me know if my comment above on why libvirt is different from openstack makes sense. As you suggest let's switch the if/elsif to a case in a follow up PR. I have some more changes coming in that area which which I can fold this in.

Anything else before this can be merged?

simaishi:
Looks good for GCE, thanks @geertj !

geertj:
@simaishi thanks!

@

geertj:
@simaishi thanks!

@Fryguy @jrafanie can this be merged please? @simaishi checked GCE and I personally checked the libvirt piece. Also the libvirt piece is trivial, it only adds the `qemu-guest-agent` package.

jrafanie:
Merged @geertj, thanks!

PR #149, opened 2016-08-22
Reduce size of image by 325MiB

geertj:
This series reduces the (compressed) size of the image by about 325MB. Prior to this patch a libvirt image was around 1.55 GB, with this patch it's reduced to 1.10GB.


geertj:
@simaishi can you have a look at this one also?

kickstarts/base.ks.erb
```diff
@@ -3,6 +3,7 @@
 logging --level=debug
 
 #version=DEVEL
+text
```
Fryguy: Does this reduce the size or was this for debugging...I thought we already had this somewhere, which is part of the reason I was asking.  @bdunne?

Fryguy:
@bdunne Please review.  I wonder if we can leverage this in the docker image as well (or we may already have done it, but I can't remember...I don't like having two separate build paths :( )

kickstarts/base.ks.erb
geertj: It's for debugging and slightly faster builds, which is useful when you're doing a lot of builds :) It does not impact the image size. Happy to separate it out in a separate patch, lmk.

kickstarts/partials/post/source_setup.ks.erb
```diff
@@ -16,7 +16,7 @@ pushd $appliance_root
 popd
 
 mkdir -p $app_root
-git clone <%= @manageiq_checkout.remote %> $vmdb_root
+git clone --separate-git-dir /tmp/manageiq.git <%= @manageiq_checkout.remote %> $vmdb_root
```
Fryguy: Keeping the git dir will full history was intentional as it allows the appliance to be used by developers as well.  If we decide to go this route, then we would need some new instructions for developers, and additionally, we should be consistent with the other git repos that we clone.

kickstarts/base.ks.erb
geertj: We didn't have this before by the way. If you build without "text", and look at the console of the VM being built (virsh list; virt-viewer <vmid>), then you see that we're running the graphical installer.

Fryguy:
I'm wondering if it's just cleaner to `rm -rf` the files or directories in question instead of doing all this tmpfs manipulation.  This way we can keep it consistent with the Docker container build which also just does an `rm -rf`, yum cache clean, etc.

geertj:
> I'm wondering if it's just cleaner to rm -rf the files or directories in question instead of doing all this tmpfs manipulation.

See the comment in this commit: https://github.com/ManageIQ/manageiq-appliance-build/pull/149/commits/98d8c4adfa823c2558fc4511fc434681cc66eedc :)

Running "rm -rf" just removes the directory entries .The data is still on disk, and will therefore be in the image.

Docker is different as a Docker layer is a file system image, not a disk image.

kickstarts/partials/post/source_setup.ks.erb
geertj: What if we made it conditional on the type of release? A nightly or test build would have it, a release build not.

kickstarts/base.ks.erb
Fryguy: Right I'm *sure* we had this previously because there were issues we couldn't handle without it.  @abellotti ?

kickstarts/partials/post/source_setup.ks.erb
Fryguy: That sounds like a good idea!

Fryguy:
@geertj Ah makes sense.  I actually remembered that about disk images, but I thought that the underlying layer was smart enough to not have it in the disk image.

kickstarts/partials/post/source_setup.ks.erb
Fryguy: @geertj If that's the case, can you please also do that with the other git repos that we clone?  (manageiq-appliance and manageiq-ui-self_service)

kickstarts/base.ks.erb
geertj: Whoops, with *before* I understood immediately before this change. It may be that we did in the past!

kickstarts/partials/post/source_setup.ks.erb
geertj: @Fryack ack on both items (also do for other repos, and do it only for release builds)

kickstarts/base.ks.erb
bdunne: This is super helpful in debugging, I know we had it at one time because @jrafanie ran into an error and this was the only way to see what it was.

Fryguy: So, there is a comment here to tell people to turn it on for debugging, but the actual line was removed: https://github.com/ManageIQ/manageiq-appliance-build/blob/d0f4d921daf92d60c48a08bd9ea722a9942f76a2/kickstarts/base.ks.erb#L2

There had to be a reason for removing it.  If not, I'm fine with this change, but I would like to make sure before adding it back in.

Fryguy: @geertj Regardless of if we keep this line or not, can you move it up next to the corresponding comment and isolate the pair of lines with blank lines...Something like this?

```text
# Kickstart file automatically generated by anaconda.

# Run the text based installer to see dependency info in screenshots
text

logging --level=debug
```

That way we can keep it in the file and comment/uncomment when needed.

kickstarts/partials/post/source_setup.ks.erb
bdunne: What if we clone with `--depth=1` then change the `.git/config` from:
`fetch = +refs/heads/master:refs/remotes/origin/master`
to:
`fetch = +refs/heads/*:refs/remotes/origin/*`

kickstarts/base.ks.erb
bdunne: I don't see any downside to running in text mode.  You can't even see any errors in graphical mode because you have to click something to display it and we usually just get a screenshot at the end.

kickstarts/partials/post/source_setup.ks.erb
Fryguy: @bdunne Will that refetch all of the history?

bdunne: Yes, the next time you fetch, it will fetch everything.  But in the short term you save ~100M

geertj:
@Fryguy, updated the text mode line as per your suggestion. Given that @bdunne didn't see any issue with always running a text mode build, text mode is now always enabled.

@Fryguy, @bdunne I did not implement the `git clone --depth 1` suggestion. I tried it, and it doesn't allow you to check out a specific commit like we do, you can only checkout the master. There may be a way to make it work by doing a shallow checkout specific to a certain rev, but I haven't investigated that. In any case I would have suggested to fully exclude the `.git` dirs from releases. A small download size is an important usability factor, and I want to get it as small as possible. Once we figure out how to do `--depth 1` related to a certain commit, we can use that for the *nightly* builds. I plan to look for other ways on top of this to further reduce the image size.

Please let me know if there is anything else.

kickstarts/base.ks.erb
```diff
@@ -1,5 +1,8 @@
 # Kickstart file automatically generated by anaconda.
+
 # Run the text based installer to see dependency info in screenshots
+text
```
jrafanie: Do we need this change?  We do the GUI install as of 7eabb6206f286 because it's easier if we need to debug a failed build and VNC into the failed build.

config/base.json
```diff
@@ -1,3 +1,4 @@
 {
-  "generate_icicle": false
+  "generate_icicle": false,
+  "oz_overrides" : "{'libvirt': {'memory': 2048}}"
```
jrafanie: What does this do?

kickstarts/partials/packages/excludes.ks.erb
```diff
@@ -9,15 +9,23 @@
...
 -iwl5000-firmware
 -iwl5150-firmware
 -iwl6000-firmware
 -iwl6000g2a-firmware
+-iwl6000g2b-firmware
 -iwl6050-firmware
-#kernel-firmware
+-iwl7260-firmware
+-iwl7265-firmware
+-kernel-firmware
```
jrafanie: How do we know it's ok to exclude these firmwares?

kickstarts/partials/post/ruby_install.ks.erb
```diff
@@ -1,15 +1,23 @@
-wget -O ruby-install.tar.gz https://github.com/postmodern/ruby-install/archive/v0.5.0.tar.gz
-tar -xzvf ruby-install.tar.gz
-rm -f ruby-install.tar.gz
-cd ruby-install*/
+curl -L https://github.com/postmodern/ruby-install/archive/v0.5.0.tar.gz | tar xvfzC - /tmp
+
+pushd /tmp/ruby-install*/
```
jrafanie: 👍

kickstarts/partials/post/ruby_install.ks.erb
```diff
@@ -1,15 +1,23 @@
...
 make install
+popd
+
+# The ruby-install will result in about 350MiB of files in /usr/local/src.
+# These do not need to be in the image, so put them on a tmpfs.
+mkdir -p /usr/local/src
+mount -t tmpfs tmpfs /usr/local/src
```
jrafanie: 👍

kickstarts/partials/post/source_setup.ks.erb
```diff
@@ -9,22 +9,25 @@ ssui_root="$repos_root/manageiq-ui-self_service"
 
 rm -rf $app_root
 
-git clone <%= @appliance_checkout.remote %> $appliance_root
+git clone <%= "--separate-git-dir /tmp/appliance.git" if @build_type == "release" %> \
+        <%= @appliance_checkout.remote %> $appliance_root
```
jrafanie: Many of us rely on the ability to git pull on the appliance's git repos or to checkout a git sha/tag/branch.  Will this be possible if the /tmp directory is gone after reboot?

Fryguy: As [discussed above](https://github.com/ManageIQ/manageiq-appliance-build/pull/149#discussion-diff-75916773) we were thinking that nightlies would have the .git directory, but release builds would not.  Does that sound ok to you or do you think that would cause a problem.

jrafanie: Thanks, I didn't see the hidden comments.

I honestly don't know if people use git on the current release builds to know if killing the .git directory would be a problem.  Maybe @geertj could make an announcement on talk.manageiq.org that future release builds will not be git enabled.  Do we start this with the next darga build or the E release?

I don't really have much need for git enabled release builds but this could affect how we(I) suggest people [apply a bug fix using patch instead of git am](http://talk.manageiq.org/t/vcenter-refresh-issues-in-rc2/650/13).  I don't even know if patch is installed by default.  I would assume so.

bdunne: Isn't git used to upgrade to the next release? i.e. darga1, darga2, etc.

jrafanie: Good point @bdunne.  It looks like it's been asked about a few times. It was asked about [here](http://talk.manageiq.org/t/miq-appliance-upgrade/755/12), [here](http://talk.manageiq.org/t/upgrading-from-darga-n-to-darga-n-1/1574), and finally using `git` and a followup to use export/import db into new appliance was advised [here](http://talk.manageiq.org/t/upgrade-to-darga/1520/6).  Either way, we should have a clear way to do upgrades... if git isn't used for upgrades, then, I'd be ok with removing the `.git` after we announce it.

geertj:
Just a quick not to all that my weekend time has dried up but that I will be getting back to this PR soon.

kickstarts/partials/post/source_setup.ks.erb
Fryguy: Oh yeah...I forgot about release upgrades... hmmm

geertj: @Fryguy @bdunne @jrafanie So it appears that we can't easily remove the .git directory. Building on @bdunne's suggestion, the following appears to work:

```
  git clone --depth 1 --branch <branch_or_tag> https://github.com/manageiq/manageiq.git
  git config remote.origin.fetch '+refs/heads/*:refs/remote/origin/*'
  git config branch.<branch>.remote origin
  git config branch.<branch>.merge refs/heads/<branch>
  git checkout -b <branch>
```

As far as I can see the result is exactly similar to the current situation (e.g. `git pull` will do the right thing automatically). The difference is about 100MB which is a good reduction.

kickstarts/partials/packages/excludes.ks.erb
geertj: I looked at all the hardware that is supposed by these firmwares (including kernel-firmware), and there isn't a hypervisor that implements the hardware as a virtual device. I checked both ESX and QEmu.

Can I guarantee that there isn't an exotic situation where these firmwares are needed in a virtual appliance? No... But IMHO it's very unlikely, and therefore I think these should be removed. In the unlikely event we'd presumable get a good error message from the kernel driver and we can put it back.

config/base.json
geertj: This increases the memory of the VM that is used by ImageFactory to create the image to from 1GB to 2GB. I did this for safety reasons since we're using a couple of tmpfs file systems in that VM. I don't have the total amount of space that they use but it is less that 1GB.

geertj:
I will update the PR shortly.

kickstarts/partials/post/source_setup.ks.erb
Fryguy: Nice @geertj ... I like this compromised approach the best so far.

jrafanie: I'm good with this as long as a darga build will be able to git pull darga updates, master git pull will get master updates, etc. 👍

kickstarts/partials/packages/excludes.ks.erb
jrafanie: I'm ok with this, but I have even less knowledge of these firmware, so I'm cautious about changing them.  I'm ok with changing this but since you dug into this a bit, if there was a missing firmware do you have any idea what step would fail in the build process?

kickstarts/base.ks.erb
jrafanie: @geertj Is this change still required?  See above note about VNC.  If we can't do VNC in text mode or the info we get in text mode VNC is less useful, I'd be against doing this.

geertj: @jrafanie you can certainly use VNC to connect to a VM that is in text mode. The VNC server is provided by the hypervisor not the guest.

kickstarts/partials/packages/excludes.ks.erb
geertj: @jrafanie if a firmware was needed, the appliance when it starts up would not be able to correctly use a virtual device. If the device was a network card, the appliance would no be able to connect to the network for example. The build would not fail.

But note that these firmwares are for Wifi cards, Fibrechannel adapters, etc. VMs never have these kinds of devices. All network and disk devices for hypervisors that we support are simple (virtio, ide, etc), and they are all supported out of the box without any firmware.

kickstarts/base.ks.erb
jrafanie: @geertj how does this impact image size?  I vaguely recall the text mode not having the ability to show some of the error/details for some failed builds.  Maybe @abellotti remembers.  We changed it to GUI installer for a reason.  If this change has minimal impact on the image size, I wouldn't want to do it.

abellotti: GUI mode was easier to monitor/view/debug/continuing interaction with installer.  (was living in that mode for a while going from 6.x to 7.1).  So if it's not too expensive image size wise, I'd prefer to keep it.  Thanks.

kickstarts/base.ks.erb
```diff
@@ -1,5 +1,8 @@
 # Kickstart file automatically generated by anaconda.
-# Run the text based installer to see dependency info in screenshots
+
+# By default run the text based installer to see dependency info in screenshots
+<% if ENV['KS_GUI'] == '1' then "" else "text" end %>
+
```
geertj: Allowing $KS_GUI=1 to move back to text mode, as requested by @abellotti.

I spoke to Alberto, and the GUI mode was useful early on when initially creating the kickstart file. If you failed to set a required option, it will prompt you for the missing value, which is easier in GUI mode. Now that we have a working kickstart and making only incremental changes, I believe text is more useful because it gives you a log of action that you can scroll up. GUI mode only shows you the current package that is being installed.

jrafanie: @geertj Thanks, that makes sense.  I'm good with this.  How much does this impact the image size though?

abellotti: Hi @geertj thanks for making this change, instead of ENV variable, I'd prefer if this was a vmbuild.rb option (as defined in cli.rb), introducing --text  and --graphics options.  If you prefer we can do this in another PR.

jrafanie:
@geertj after some of the settings were changed a little (namely how we checkout the repo with git), etc., is the name of this PR still accurate:  `Reduce size of image by 450MB `?

kickstarts/base.ks.erb
jrafanie: @abellotti I agree.  Let's do this in a followup PR.

jrafanie:
LGTM, @simaishi can you confirm you're good with this?  Thanks.

geertj:
@jrafanie it's 325MiB now. I've updated the title.

kickstarts/base.ks.erb
geertj: @jrafanie there is no impact (positive or negative) on image size. The benefit of doing a text mode install is a slightly faster install (2 mins) and the fact that you see a log of what's happening on the screen.

kickstarts/partials/packages/excludes.ks.erb
```diff
@@ -1,33 +1,18 @@
...
+-alsa-*
+
+# We don't need any firmware files. We always run as a virtual appliance, and
+# none of the virtual hardware devices implement by any of the major hypervisor
+# vendors needs firmware.
+-*-firmware
+
+# NetworkManager is not needed. The appliance uses the "traditiona" way of
+# networking.
+-NetworkManager*
```
simaishi: If we'll be excluding NetworkManager, we should probably update systemd.ks.erb:

https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/systemd.ks.erb#L20

kickstarts/partials/packages/excludes.ks.erb
```diff
@@ -1,33 +1,18 @@
...
+
+# ALSA is not needed. We don't need to play sounds in our appliance.
+-alsa-*
+
+# We don't need any firmware files. We always run as a virtual appliance, and
+# none of the virtual hardware devices implement by any of the major hypervisor
+# vendors needs firmware.
+-*-firmware
+
+# NetworkManager is not needed. The appliance uses the "traditiona" way of
```
jrafanie: Thanks @simaishi, now I see a typo:  `traditional`

kickstarts/partials/packages/excludes.ks.erb
jrafanie: @carbonin is it ok to just not install `NetworkManager`?  I believe you added the line to disable networker manager (above).

kickstarts/partials/packages/excludes.ks.erb
geertj: Typo fixed.

kickstarts/partials/packages/excludes.ks.erb
geertj: @simaishi removed "systemctl disable NetworkManager" line from systemd.ks.erb

kickstarts/partials/packages/excludes.ks.erb
```diff
@@ -1,33 +1,18 @@
...
+-alsa-*
+
+# We don't need any firmware files. We always run as a virtual appliance, and
+# none of the virtual hardware devices implement by any of the major hypervisor
+# vendors needs firmware.
+-*-firmware
+
+# NetworkManager is not needed. The appliance uses the "traditional" way of
+# networking.
+-NetworkManager*
```
abellotti: installed but disabled.  some customers may enable it to do realm discovery for AD domains before joining them (external auth AD).

geertj: @abellotti OK so we need to revert this change?

geertj: done

geertj:
Excellent. Thanks all!

Fryguy:
This will conflict, so backport of the size changes is in #157

PR #153, opened 2016-09-07
Fix when building from tag

geertj:
Old code (all run as subshells):

  git reset --hard
  git clean -dxf
  git checkout <ref>
  git fetch origin
  git reset --hard origin/<ref>

The last line only works when the --reference <ref> parameter is a
branch. The build does not fail though, because a subshell failing
doesn't exit the Ruby interpreter. So the code was in essence a NOP in
case of a branch.

However, if I understand the old code correctly, there is another issue.
If you build from a new tag for the first time, the code would actually
not build that tag but whatever was there before. This is because the
"git fetch" is done after the "git checkout <ref>". This could actually
have lead to releases that don't correspond to their tag :-D

The new code looks like this:

  rm -rf /build/references/<ref>
  chdir /build/references/<ref>
  git clone <url> .
  git checkout <ref>

The "rm -rf" should be innocuous. It will never remove your working repo
which is at /build/manageiq-appliance-build. It will only remove a repo
that the script itself checked out. It's safe even if you forget --local.

### DIFF
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -46,18 +46,12 @@ FILE_SERVER_BASE    = Pathname.new(ENV["BUILD_FILE_SERVER_BASE"] || ".") # Subdi
 if !cli_options[:local] && cli_options[:build_url]
   build_repo = cli_options[:build_url]
   cfg_base = REFS_DIR.join(cli_options[:build_ref])
+  FileUtils.rm_rf(cfg_base)
   FileUtils.mkdir_p(cfg_base)
   Dir.chdir(cfg_base) do
-    unless File.exist?(".git")
-      $log.info("Cloning Repo #{build_repo} to #{cfg_base} ...")
-      `git clone #{build_repo} .` unless File.exist?(".git")
-    end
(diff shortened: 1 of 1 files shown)


simaishi:
@geertj with the old code, building from a new tag always worked as it created a new directory with that new tag name. Which is probably why the last line didn't matter when doing a tag build for our official releases.

geertj:
@simaishi ah!! Indeed a new tag would cause a fresh checkout. So builds were OK.

Nevertheless the fact that the code is working depends on two side effects (reset being an ignored error for tags, and creating a new directory for a new tag) , which makes it quite fragile, and hard to understand. I think this patch simplifies things.

abellotti:
How much longer does it take to do the git clone route vs. old code ?

Probably negligible considering the whole build time, but would be good to know.

geertj:
@abellotti Just a few seconds. The manageiq-appliance-build repo is very small.

jrafanie:
@abellotti LGTM, it's certainly inefficient to blow away the directory and re-clone but it's 
* much less steps
* guaranteed to clean things up
* a tiny repo so git clone is not expensive

Please merge if you're good to go on this one.

abellotti:
👍

PR #154, opened 2016-09-08
[WIP] vmpublish.rb: publish public images

geertj:
This script publishes builds from a previous vmbuild.rb invocation to
any of the configurated public directories. Currently this means Atlas
(for Vagrant images) and GCE (for GCE images).

This code could probably use a similar metadata driven approach as
vmbuild.rb. However for now I believe the current approach of hardcoding
some of the calling conventions and parameters for each of the supported
targets works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /productization
 kickstarts/generated/*.ks
 scripts/Gemfile.dev.rb
+config/creds/*.json
--- a/bin/release-build.sh
+++ b/bin/release-build.sh
@@ -12,5 +12,10 @@ if [[ $# != 1 ]]; then
 fi
 
 log_file="/build/logs/${1}.log"
-nohup time ruby ${BUILD_DIR}/scripts/vmbuild.rb --type release --upload --reference $1 --copy-dir ${1%%-*} > $log_file 2>&1 &
+copy_dir="${1%%-*}"
+
+nohup sh -c \
+        "echo time ruby ${BUILD_DIR}/scripts/vmbuild.rb --type release --upload --reference $1 --copy-dir $copy_dir \
+         && time ruby ${BUILD_DIR}/scripts/vmpublish.rb --copy-dir $copy_dir" \
+        > $log_file 2>&1 &
 echo "${1} release build kicked off, see log @ $log_file ..."
--- a/config/creds/README.creds
+++ b/config/creds/README.creds
@@ -1,0 +1,21 @@
+This directory will hold credentials files on the build host.
+
+All JSON files in this directory are ignored via gitignore(5) to reduce the
+risk that credentials accentally get commit.
+
+The following files are currently used:
+
+* gce.json
+
+  A service account JSON file for the Google Cloud Platform. This file can be
+  generated as describe here:
+  https://cloud.google.com/storage/docs/authentication#generating-a-private-key
+
+* atlas.json
+
+  A JSON file with an Atlas token, as follows:
+
+  {
+      "username": "<username>",
+      "token": "<token"
+  }
--- a/scripts/config.rb
+++ b/scripts/config.rb
@@ -1,0 +1,15 @@
+require 'pathname'
+
+BUILD_BASE          = Pathname.new("/build")
+GPG_DIR             = Pathname.new("/root/.gnupg")
+CFG_DIR             = BUILD_BASE.join("config")
+CREDS_DIR           = CFG_DIR.join("creds")
+FILESHARE_DIR       = BUILD_BASE.join("fileshare")
+REFS_DIR            = BUILD_BASE.join("references")
+IMGFAC_DIR          = BUILD_BASE.join("imagefactory")
+IMGFAC_CONF         = CFG_DIR.join("imagefactory.conf")
+STORAGE_DIR         = BUILD_BASE.join("storage")
+
+FILE_SERVER         = ENV["BUILD_FILE_SERVER"]             # SSH Server to host files
+FILE_SERVER_ACCOUNT = ENV["BUILD_FILE_SERVER_ACCOUNT"]     # Account to SSH as
+FILE_SERVER_BASE    = Pathname.new(ENV["BUILD_FILE_SERVER_BASE"] || ".") # Subdirectory of Account where to store builds
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -10,6 +10,7 @@ require_relative 'git_checkout'
 require_relative 'cli'
 require_relative 'uploader'
 require_relative 'target'
+require_relative 'config'
 
 $log = Logger.new(STDOUT)
 
@@ -29,19 +30,6 @@ else
   puddle = cli_options[:type]
   build_label = "#{cli_options[:type]}-#{cli_options[:manageiq_ref]}"
 end
-
-BUILD_BASE          = Pathname.new("/build")
-GPG_DIR             = Pathname.new("/root/.gnupg")
-CFG_DIR             = BUILD_BASE.join("config")
-FILESHARE_DIR       = BUILD_BASE.join("fileshare")
-REFS_DIR            = BUILD_BASE.join("references")
-IMGFAC_DIR          = BUILD_BASE.join("imagefactory")
-IMGFAC_CONF         = CFG_DIR.join("imagefactory.conf")
-STORAGE_DIR         = BUILD_BASE.join("storage")
-
-FILE_SERVER         = ENV["BUILD_FILE_SERVER"]             # SSH Server to host files
-FILE_SERVER_ACCOUNT = ENV["BUILD_FILE_SERVER_ACCOUNT"]     # Account to SSH as
-FILE_SERVER_BASE    = Pathname.new(ENV["BUILD_FILE_SERVER_BASE"] || ".") # Subdirectory of Account where to store builds
 
 if !cli_options[:local] && cli_options[:build_url]
   build_repo = cli_options[:build_url]
@@ -101,8 +89,7 @@ ks_gen.run
 file_rdu_dir_base = FILE_SERVER_BASE.join(directory)
 file_rdu_dir      = file_rdu_dir_base.join(directory_name)
 
-fileshare_dir         = BUILD_BASE.join("fileshare")
-stream_directory      = fileshare_dir.join(directory)
+stream_directory      = FILESHARE_DIR.join(directory)
 destination_directory = stream_directory.join(build_label == "test" ? "test" : directory_name)
 
 $log.info "Creating Fileshare Directory: #{destination_directory}"
@@ -152,7 +139,7 @@ Dir.chdir(IMGFAC_DIR) do
     FileUtils.mkdir_p(destination_directory)
     file_name = "#{name}-#{target}-#{build_label}-#{timestamp}-#{manageiq_checkout.commit_sha}.#{target.file_extension}"
     destination = destination_directory.join(file_name)
-    $log.info `mv #{source} #{destination}`
+    $log.info `ln #{source} #{destination}`
 
     if !File.exist?(destination)
       $log.warn "Cannot find the target file #{destination}"
--- a/scripts/vmpublish.rb
+++ b/scripts/vmpublish.rb
@@ -1,0 +1,80 @@
+require 'logger'
+require 'pathname'
+require 'trollop'
+require 'json'
+
+require_relative 'config'
+
+
+def find_imgfac_image(image)
+  f = File.new(image)
+  Dir.chdir(STORAGE_DIR) do
+    Dir.glob('*.body') do |filename|
+      return filename.split('.').first if File.new(filename).stat.ino == f.stat.ino
+    end
+  end
+end
+
+
+def upload_imgfac_image(uuid, target, reference, version)
+  if target == 'vagrant'
+    creds_file = CREDS_DIR.join('atlas.json')
+    params = "atlas @ignored #{creds_file} --id #{uuid}"
+    params += " --parameter atlas_box_name #{reference} --parameter atlas_box_version #{version}"
+  elsif target == 'gce'
+    creds_file = CREDS_DIR.join('gce.json')
+    params = "gce @manageiq #{creds_file} --id #{uuid}"
+    params += " --parameter gce_object_name manageiq-#{reference}-#{version}.tar.gz"
+    params += " --parameter gce_image_name manageiq-#{reference}-#{version} --parameter gce_image_family centos-7"
+  end
+
+  if !File.exist?(creds_file)
+    $log.info "credentials file #{creds_file} does not exist, skipping #{target} publish"
+    return
+  end
+
+  Dir.chdir(IMGFAC_DIR) do
+    # Show progress on STDOUT
+    command = "./imagefactory provider_image #{params}"
+    $log.info "running command: #{command}"
+    if !system(command)
+      $log.error "imagefactory exited with status #{$?.exitstatus}"
+      exit 1
+    end
+  end
+end
+
+
+$log = Logger.new(STDOUT)
+
+dir_desc  = "Directory builds were copied to"
+type_desc = "Build type (stable or latest)"
+
+options = Trollop.options(ARGV) do
+  banner "Usage: vmpublish.rb [options]"
+  opt :copy_dir,  dir_desc,   :type => :string,   :short => 'd', :default => "master"
+  opt :type,      type_desc,  :type => :string,   :short => 't', :default => "stable"
+end
+
+copy_dir = FILESHARE_DIR.join(options[:copy_dir], options[:type])
+
+Dir.foreach(copy_dir) do |filename|
+  next if ['.', '..'].include? filename
+
+  parts = filename.split('.').first.split('-')
+  next unless parts.length >= 6        # only release builds, which are build from a tag (name-version)
+  next unless parts[0] == 'manageiq'
+  target = parts[1]
+  next unless ['vagrant', 'gce'].include? target
+  reference = parts[2]
+  version = parts[3..parts.length-3].join('-')
+
+  image = copy_dir.join(filename)
+  uuid = find_imgfac_image(image)
+
+  $log.info "processing image: #{image}"
+  $log.info "target = #{target}, reference = #{reference}, version = #{version}"
+  $log.info "imagefactory image = #{uuid}"
+
+  upload_imgfac_image(uuid, target, reference, version)
+end


scripts/vmbuild.rb
```diff
@@ -30,19 +31,6 @@
...
-FILESHARE_DIR       = BUILD_BASE.join("fileshare")
-REFS_DIR            = BUILD_BASE.join("references")
-IMGFAC_DIR          = BUILD_BASE.join("imagefactory")
-IMGFAC_CONF         = CFG_DIR.join("imagefactory.conf")
-STORAGE_DIR         = BUILD_BASE.join("storage")
-
-FILE_SERVER         = ENV["BUILD_FILE_SERVER"]             # SSH Server to host files
-FILE_SERVER_ACCOUNT = ENV["BUILD_FILE_SERVER_ACCOUNT"]     # Account to SSH as
-FILE_SERVER_BASE    = Pathname.new(ENV["BUILD_FILE_SERVER_BASE"] || ".") # Subdirectory of Account where to store builds
-
```
jrafanie: @geertj from a reviewer perspective, I have to scroll up and down to see if this code was just moved or moved and modified.  It looks identical, but moved.  Can you do the move to a new file in a single commit so I can review commit by commit and not have to keep all of these changes in my head?

scripts/config.rb
```diff
@@ -0,0 +1,15 @@
...
+CREDS_DIR           = CFG_DIR.join("creds")
+FILESHARE_DIR       = BUILD_BASE.join("fileshare")
+REFS_DIR            = BUILD_BASE.join("references")
+IMGFAC_DIR          = BUILD_BASE.join("imagefactory")
+IMGFAC_CONF         = CFG_DIR.join("imagefactory.conf")
+STORAGE_DIR         = BUILD_BASE.join("storage")
+
+FILE_SERVER         = ENV["BUILD_FILE_SERVER"]             # SSH Server to host files
+FILE_SERVER_ACCOUNT = ENV["BUILD_FILE_SERVER_ACCOUNT"]     # Account to SSH as
+FILE_SERVER_BASE    = Pathname.new(ENV["BUILD_FILE_SERVER_BASE"] || ".") # Subdirectory of Account where to store builds
```
jrafanie: So, we need to put these constants within the `Build` namespace.  Meaning, they should reside in `module Build`.  We can do this later but we really should have them in that namespace and all code that uses it, if it's in that namespace, can access it without needing to specify the namespace, meaning we can use `BUILD_BASE` within the namespace instead of  `Build::BUILD_BASE`.  When we put it in the `Build` namespace, it needs to live in a build directory, so `scripts/build/config.rb`...

Because it's relevant here but not required... I created #158 for the following comment:

Future PR: Note, we really need to move files to the right places and re-namespace things...

```
vmbuild.rb             -> build.rb
cli.rb                 -> build/cli.rb
git_checkout.rb        -> build/git_checkout.rb
kickstart_generator.rb -> build/kickstart_generator.rb
productization.rb      -> build/productization.rb
target.rb              -> build/target.rb
productization.rb      -> build/productization.rb
uploader.rb            -> build/uploader.rb
```

Also, rename scripts to lib.

.gitignore
```diff
@@ -2,3 +2,4 @@
 /productization
 kickstarts/generated/*.ks
 scripts/Gemfile.dev.rb
+config/creds/*.json
```
jrafanie: 👍

bin/release-build.sh
```diff
@@ -12,5 +12,10 @@ if [[ $# != 1 ]]; then
 fi
 
 log_file="/build/logs/${1}.log"
-nohup time ruby ${BUILD_DIR}/scripts/vmbuild.rb --type release --upload --reference $1 --copy-dir ${1%%-*} > $log_file 2>&1 &
+copy_dir="${1%%-*}"
+
+nohup sh -c \
+        "echo time ruby ${BUILD_DIR}/scripts/vmbuild.rb --type release --upload --reference $1 --copy-dir $copy_dir \
+         && time ruby ${BUILD_DIR}/scripts/vmpublish.rb --copy-dir $copy_dir" \
```
jrafanie: Is publishing a step in the build process?  It feels like it is.  That's why it feels odd to me that we need to do this multiline shell command.  Can we instead change vmbuild.rb to accept a `--publish` option with any other details we need?

config/creds/README.creds
```diff
@@ -0,0 +1,21 @@
+This directory will hold credentials files on the build host.
+
+All JSON files in this directory are ignored via gitignore(5) to reduce the
+risk that credentials accentally get commit.
```
jrafanie: `get committed`

config/creds/README.creds
```diff
@@ -0,0 +1,21 @@
...
+
+The following files are currently used:
+
+* gce.json
+
+  A service account JSON file for the Google Cloud Platform. This file can be
+  generated as describe here:
```
jrafanie: `described`

config/creds/README.creds
```diff
@@ -0,0 +1,21 @@
...
+  https://cloud.google.com/storage/docs/authentication#generating-a-private-key
```
jrafanie: Is there something we need to install on the build environment to talk to google here?  Maybe this needs to be added to the base README.

config/creds/README.creds
jrafanie: We call it `gce.json` but say `Google Cloud Platform`.  Which one should it be for consistency?

config/creds/README.creds
```diff
@@ -0,0 +1,21 @@
...
+
+* atlas.json
+
+  A JSON file with an Atlas token, as follows:
```
jrafanie: Same as above, what do we need to install and can we update the README with that information?

bin/release-build.sh
simaishi: I agree.. just like how `--upload` is done.

scripts/vmbuild.rb
```diff
@@ -152,7 +139,7 @@ def verify_run(output)
     FileUtils.mkdir_p(destination_directory)
     file_name = "#{name}-#{target}-#{build_label}-#{timestamp}-#{manageiq_checkout.commit_sha}.#{target.file_extension}"
     destination = destination_directory.join(file_name)
-    $log.info `mv #{source} #{destination}`
+    $log.info `ln #{source} #{destination}`
```
jrafanie: what's this change for?

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
+require 'logger'
+require 'pathname'
+require 'trollop'
+require 'json'
+
+require_relative 'config'
+
+
```
jrafanie: This extra whitespace here indicates in my mind that the methods below belong in a namespace, such as `Build` if publishing is a step in the build process.

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+def find_imgfac_image(image)
```
jrafanie: Do we need this method if the build process "hands" over the image to the publishing step directly?

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+      return filename.split('.').first if File.new(filename).stat.ino == f.stat.ino
+    end
+  end
+end
+
+
+def upload_imgfac_image(uuid, target, reference, version)
+  if target == 'vagrant'
+    creds_file = CREDS_DIR.join('atlas.json')
+    params = "atlas @ignored #{creds_file} --id #{uuid}"
```
jrafanie: As mentioned above, we'll need instructions on installing atlas in the README, same for gce below.

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+  Dir.chdir(STORAGE_DIR) do
+    Dir.glob('*.body') do |filename|
+      return filename.split('.').first if File.new(filename).stat.ino == f.stat.ino
+    end
+  end
+end
+
+
+def upload_imgfac_image(uuid, target, reference, version)
+  if target == 'vagrant'
```
jrafanie: This conditional might we need to be refactored in the future.  It's fine for now though.

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+    params = "atlas @ignored #{creds_file} --id #{uuid}"
+    params += " --parameter atlas_box_name #{reference} --parameter atlas_box_version #{version}"
+  elsif target == 'gce'
+    creds_file = CREDS_DIR.join('gce.json')
+    params = "gce @manageiq #{creds_file} --id #{uuid}"
+    params += " --parameter gce_object_name manageiq-#{reference}-#{version}.tar.gz"
+    params += " --parameter gce_image_name manageiq-#{reference}-#{version} --parameter gce_image_family centos-7"
+  end
+
+  if !File.exist?(creds_file)
```
jrafanie: Should this check happen earlier, before `atlas` and `gce` commands?  Otherwise, if the file doesn't exist, we'll try to run those commands with a missing file.

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+def upload_imgfac_image(uuid, target, reference, version)
+  if target == 'vagrant'
+    creds_file = CREDS_DIR.join('atlas.json')
+    params = "atlas @ignored #{creds_file} --id #{uuid}"
+    params += " --parameter atlas_box_name #{reference} --parameter atlas_box_version #{version}"
+  elsif target == 'gce'
+    creds_file = CREDS_DIR.join('gce.json')
+    params = "gce @manageiq #{creds_file} --id #{uuid}"
+    params += " --parameter gce_object_name manageiq-#{reference}-#{version}.tar.gz"
+    params += " --parameter gce_image_name manageiq-#{reference}-#{version} --parameter gce_image_family centos-7"
```
jrafanie: We might want to use AwesomeSpawn to build these comandline.

See [run and run! here](https://github.com/ManageIQ/awesome_spawn/blob/db556875484e296d940d515863434c1f964f2ebf/lib/awesome_spawn.rb#L19-L69)

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+    $log.info "credentials file #{creds_file} does not exist, skipping #{target} publish"
+    return
+  end
+
+  Dir.chdir(IMGFAC_DIR) do
+    # Show progress on STDOUT
+    command = "./imagefactory provider_image #{params}"
+    $log.info "running command: #{command}"
+    if !system(command)
+      $log.error "imagefactory exited with status #{$?.exitstatus}"
```
jrafanie: Same as above, we should use AwesomeSpawn run, and look at the standard out and standard error for logging.  Here, we don't log any errors.

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+
+$log = Logger.new(STDOUT)
+
+dir_desc  = "Directory builds were copied to"
+type_desc = "Build type (stable or latest)"
+
+options = Trollop.options(ARGV) do
+  banner "Usage: vmpublish.rb [options]"
+  opt :copy_dir,  dir_desc,   :type => :string,   :short => 'd', :default => "master"
+  opt :type,      type_desc,  :type => :string,   :short => 't', :default => "stable"
```
jrafanie: Same question as above, does publish belong as a step in the build process, if so, we should move the ARGV options there.

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
...
+Dir.foreach(copy_dir) do |filename|
+  next if ['.', '..'].include? filename
+
+  parts = filename.split('.').first.split('-')
+  next unless parts.length >= 6        # only release builds, which are build from a tag (name-version)
+  next unless parts[0] == 'manageiq'
+  target = parts[1]
+  next unless ['vagrant', 'gce'].include? target
+  reference = parts[2]
+  version = parts[3..parts.length-3].join('-')
```
jrafanie: I'm not sure what this is doing.

scripts/vmpublish.rb
jrafanie: Note, I don't think we use AwesomeSpawn yet in this repo so maybe we shouldn't.  cc @carbonin @simaishi thoughts?

bin/release-build.sh
geertj: I moved it to a separate segment on purpose. Happy to reconsider that but let me explain my reasoning here first to see if it make sense.

These were my thoughts:

1. There can be multiple locations to which you upload a single image. Once we support AWS we will get into this situation, where we have to upload the same image to multiple regions. We'd have to have a nested loop in the build loop and I though that would be more complicated and make things really slow.

2. It allows some sanity tests before publishing. I  In my view "publish" is a more impactful operation than "upload". The upload puts stuff on our file server from which people can download it. But "publish" will immediate make the image the live default choice for users of those services. For example, if you use "vagrant init manageiq/darga" that would immediately take the latest version by default. Same for the cloud launcher on GCE (we are working on this now). So doing some quick sanity testing on the build before a publish would not be a bad thing.

(Note that this means we shouldn't call the publish script unconditionally from the "release-build.sh" script. That was an oversight.)

3. Separation makes things more reliable. It allows the build phase to fail, without having some publish locations updated and others not. And it also allows the publish phase to be idempotent. If the publish fails, simply re-run the command. It won't build anything, but re-upload the most recently built set of images. The way the script works, it will overwrite anything with the specified version that is already there.

config/creds/README.creds
geertj: ImageFactory calls it GCE so probably that's the name to use.

config/creds/README.creds
geertj: The Google API client is needed for GCE, which is a transitive dependency of the new ImageFactory and documented here:

https://github.com/redhat-imaging/imagefactory/blob/master/imagefactory_plugins/GCE/README.gce

I think our README can be cleaned up. Some of the dependencies that are mentioned are not needed anymore in recent imagefactory. I got some PRs accepted that turn some hard dependencies soft dependencies (which we don't need/use).

What if we agree that I'll simplify the README in a follow up PR (shortly, ~few weeks), would that be OK?

config/creds/README.creds
geertj: As mentioned above this is all done in imagefactory. I'll get the README updates shortly.

scripts/vmbuild.rb
geertj: I need the imagefactory image to continue to exist in /build/storage, so that I can imagefactory to publish the image in a separate step (becomes moot if we collapse the steps). I'm using a hard link instead of a copy so that I can more efficiently find back the image UUID from the image in /build/fileshare.

scripts/vmpublish.rb
geertj: Not needed if we publish in the inner build loop. See the comments on why I separated the steps.

scripts/vmpublish.rb
geertj: ack

scripts/vmpublish.rb
geertj: `system` runs the command with stdout/err connected to the terminal. I chose that on purpose because imagefactory provides progress information on stdout (these uploads can take a while) and I wanted to know who things were going during development. Also stderr would show on the terminal (or in nohup.out). And on a nonzero exit status, I exit.

scripts/vmpublish.rb
geertj: It parses back the image file name "manageiq-{ref}-{data}-{commit}" back to its components, so that I can use these are parameters to the upload.

bin/release-build.sh
simaishi: 1. Don't we have the same problem even with the current approach?  It seems each target has a hard-coded path to a .json file and we publish once per target. 

2. As you said, that's a valid point only if publish isn't a part of the release-build.sh!

3. I think this can go either way... It might make sense that we can publish some not all.  But since we're trying to publish 'release' builds only, if a release build failed for some, we might want to not publish any until we know what the problem is?

geertj: 1. Yes, although in `vmpublish.rb` I could loop over the publish destination and not the image type.

2. I propose to remove it from there :)

3. Agreed that we don't want to publish if parts of the build is broken. I was referring to the fact that the easiest way to do the upload is to add it in the inner build loop in `vmbuild.rb` before we rename the image. That would mean we interleave things: build X - upload X - build Y, upload Y.  We could put the publish loop of course completely at the end of `vmbuild.rb`.

The strongest argument for separating the build and publish phases is 2 - so that you can do some sanity testing before you publish something to a potentially large number of users. What do you think about that?

scripts/vmpublish.rb
carbonin: We may not gain a whole lot by using AwesomeSpawn here. Because of how the parameters are listed, we will likely just have something like `:params => %w(all the params here)` which doesn't do a lot for readability. Also we don't seem to be doing much escaping here (the other reason to use it) so I feel like we could keep it as is.

bin/release-build.sh
jrafanie: I can see it both ways.  My main concern is that there is some complexity because the build code isn't handing off to the publish code so we have to find the right images, and do some things that the build code would just know.  

So, how about this... can we make the publish code separate but still offer the option to run it for all successful builds?  I think that's what this shell code change here is saying.  I'd rather the build code call publish code based on the results of each build.  We can add publish CLI code later once the interface for publish is done.

config/creds/README.creds
jrafanie: I'm good with a very short addition to the README indicating these new features require `...` that we can later cleanup and add more details.

jrafanie: We'll need these details to update our build vms anyway, but yes, that can wait as long as we have the basics.

scripts/vmbuild.rb
jrafanie: Right.  Ok.

scripts/vmpublish.rb
jrafanie: ok

scripts/vmpublish.rb
jrafanie: This might ok because I don't know that this feature should introduce AwesomeSpawn as a dependency.

scripts/vmpublish.rb
```diff
@@ -0,0 +1,80 @@
+require 'logger'
```
jrafanie: Can we name this `publish.rb`?  We really need to rename vmbuild.rb to build.rb since that file adds the `Build` module.

scripts/vmpublish.rb
jrafanie: If we keep this separate, that's fine, but this should be in a `Publish` module.

scripts/vmpublish.rb
simaishi: The files in /build/storage get cleaned up daily (if you follow our instruction to setup build machine).  If the steps are separate and `publish` can happen later, *.body files might be gone.

.gitignore
Fryguy: Why not just ignore the entire directory instead of only json files?

geertj:
Sorry for the delay guys. Let's try to get resolution on the design issue first: whether this in the first instance needs to be a separate executable program, or be part of `vmbuild.rb`.

I proposed three reasons in favor of a separate program, and some of those were rightfully challenged. But taking a step back, I think the reason behind all of these is my intuition shaped by the Unix philosophy: small utilities that do one thing and do it well, and that communicate via pipes (or files).

The benefits are that the individual program become easier to develop and debug. Errors are localized, and everything becomes easier to test (I realize we have unit tests and stuff.. but still) And we can also use the best tool for the job. Take for example #161 where I add support for uploading to EC2. I decided to use Ansible for that given that I needed a mix of API and secure shell actions. Ansible makes that very easy to do. The smaller program can then be tied together with a script if we need to.

One consequence is that we need to be less aggressive cleaning up the files in `/build/storage`. It contains essential information that is needed for the upload. See the EC2 playbook where I parse the `.meta` file to get the uncompressed file size for the image. I think if you keep files for at least 2 days you are mostly good. That gives you 2 days to do manual uploads and test things out. (And if an image gets delete that's not a problem as it can be recreated - it just takes time.)

PR #156, opened 2016-09-13
[DARGA] Add support for Vagrant

geertj:
This is a first version. Current it is limited to the VirtualBox format
(which is by far the most common format). Also uploading to the
Hashicorp Altas is not yet supported.

(Cherry picked from ca57df70bd4dbb1a50dfa0eeaca07737727d093a)

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -96,6 +96,7 @@ ramfsfile="/boot/initramfs-$kversion.img"
 <%= render_partial "post/aws-ec2" if @target == "aws-ec2" %>
 <%= render_partial "post/azure" if @target == "azure" %>
 <%= render_partial "post/gce" if @target == "gce" %>
+<%= render_partial "post/vagrant" if @target == "vagrant" %>
 
 # make sure there is a new line at the end of sshd_config
 echo "" >> /etc/ssh/sshd_config
--- a/kickstarts/partials/post/vagrant.ks.erb
+++ b/kickstarts/partials/post/vagrant.ks.erb
@@ -1,0 +1,23 @@
+# Add a user "vagrant" with password "vagrant" and a well known public key.
+# Note this is how Vagrant works. You are assumed not to use a Vagrant box on
+# a publicly acessible network.
+useradd vagrant -p b1HEow3cJo7Nc
+
+cat << EOM > /etc/sudoers.d/vagrant-nopasswd
+Defaults:vagrant !requiretty
+vagrant ALL=(ALL) NOPASSWD:ALL
+EOM
+sed -i 's/.*UseDNS.*/UseDNS no/' /etc/ssh/sshd_config
+chmod 400 /etc/sudoers.d/vagrant-nopasswd
+
+mkdir -m 0700 -p ~vagrant/.ssh
+cat << EOM > ~vagrant/.ssh/authorized_keys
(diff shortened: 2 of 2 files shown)


simaishi:
@geertj Since Vagrant isn't being added to the list of 'targets' (as it was done in https://github.com/ManageIQ/manageiq-appliance-build/pull/144), this PR alone won't be adding Vagrant support.  Is the plan to build Vagrant for Darga branch going forward?

geertj:
@simaishi with this patch you can use the master branch of manageiq-appliance-build to build the darga branch of the image.

The changes were too large (IMHO) to backport all of them. And I didn't think it was necessary I was assuming that we always use a master checkout on the builder. It has support for checking out the kickstart files from an older tag or branch specified with --references (if --local is not set).

simaishi:
We use the respective branch to build each branch build, as files in 'bin' and 'scripts' are different between branches.

Our builds are kicked off by `nightly-build.sh` or `release-build.sh` from each branch, so Vagrant will still not be built for Darga by the build machine.

geertj:
@simaishi are you saying you have multiple build VMs? (one per branch) It seemed to me builds are always built from /build/manageiq-appliance-build. Or are you checking out different versions into that directory?

simaishi:
Checking out 2 branches to 2 different directories and run `release-build.sh` or 'nightly-build.sh` from each directory.

geertj:
Ahh.. So is there a solution here? I prefer not to have to backport the changes I made to `vmbuild.rb`. I could keep the changes in a private branch and use the master vmbuild.rb to build the Darga Vagrant image there. But that would suck.

I'm quite surprised actually that we're not using the master `vmbuild.rb` to build order versions, given that the code actually supports it :) I thought this is the reason we check out a specific branch's kickstart files into `/build/references`. If not why do we do this? You could always have separate build directories (one per branch) and use `--local`. So that feature would not be needed.

geertj:
Another suggestion: even if we don't build the images by default, could we still accept this patch? It is 100% a no-op with the older builder (which never has `target = "vagrant"`). But it would allow me to build it from the main repo.

simaishi:
Right, it's a no-op PR, which is why I asked if the plan is to build Vagrant for Darga branch 😄

Fryguy:
LGTM.  @chessbyte Please merge if you are good.

Fryguy:
@geertj Bump.  If you want Vangrant built for darga, then you will need to bring back the changes to darga such that the darga branch can be used for build.  Otherwise you won't have Vagrant builds for darga-4, let alone darga-3.

geertj:
@Fryguy @simaishi I would like to propose this approach to get a Darga-4 Vagrant box:

1. We merge this PR.
2. We also merge PR #157. (I need to update the PR to refer to the correct cherry picked commits).
3. When darga-4 is released we do a once-off manual build to build the Vagrant box. Using the master branch builder, you simply invoke it as `vmbuild.rb --reference  darga-4 --type release --only vagrant`. I have tested it and it works fine.

The regular nightly builds continue use the Darga builder and they will therefore not build the Vagrant box (this PR is a NOP for the Darga builder). Starting at Euwe, we will have nightly builds for Vagrant.

If we insist we want to build a Darga Vagrant box with the Darga builder then we have to backport all of #144. There's a lot of changes to `vmbuild.rb` in there. It's a lot of work for really no gain.

Fryguy:
@geertj That sounds fine to me.

> we have to backport all of #144. There's a lot of changes to vmbuild.rb in there

In hindsight that PR should have been broken up into 2 separate PRs.  One PR to add the vagrant functionality and a separate PR to do all the consolidation and refactoring.

geertj:
> In hindsight that PR should have been broken up into 2 separate PRs. One PR to add the vagrant functionality and a separate PR to do all the consolidation and refactoring.

@Fryguy happy to do that next time. Note that #144 did break out the refactoring from the vagrant support (which requires the refactoring) into two separate commits. But both commits were in the same PR. So just confirming that 1 commit per PR is preferred?

Fryguy:
In this case, yeah.  Then we could have backported the vagrant support and left behind the refactoring.

PR #160, opened 2016-09-21
Fix bug when setting KS_GUI=1

geertj:
Thanks to @Fryguy for educating me on ERB :+1:


chessbyte:
@Fryguy This did not make it into Euwe branch. Please label euwe/yes or euwe/no, as appropriate.

PR #161, opened 2016-09-22
Add support for building EC2 images.

geertj:
Note to @simaishi, the target name `'aws-ec2'` worsened our already inconsistent naming convention by adding a third one :) Now we have three:
- Named after the cloud, e.g azure.
- Named after the compute service e.g. gce.
- Named after cloud and compute with hyphen: aws-ec2

It would be nice to get some consistency here. Any suggestions?


geertj:
Two further comments:

To build this you need redhat-imaging/imagefactory#385 which is not merged yet.

To test it, for now you need to manually upload the image to EBS. You need to create an instance, attach an EBS volume to it, and dd the image on it. Then create a snapshot, and from the snapshot you create an image. I have tested it and it appears to work.

We will need automation around uploading this to Amazon as well. My current thought is that I probably wouldn't use imagefactory for this. The EC2 "provider_image" code in imgfactory is hairy and quite invasive in the changes it makes to the appliance. It is probably easier to write a simple Ansible playbook to do it. One of the benefits is that Ansible supports the CopyImage call, so that you have to create an AMI in one region only, and can then efficiently copy it to the other regions.

geertj:
Fixed the tests. Are the tests that check for the list of target names really useful?

simaishi:
@geertj We probably just used the names referred when those tasks came to us, without thinking much.

I think `<cloud>-<service>` makes most sense if we want to make them consistent. But I've never heard anyone using anything but "Azure" for Azure...  I hear people saying "Google Cloud".  I'm not familiar with all the services each cloud has nor what's commonly used in real world.

This is more of a marketing question than technical - I'm open to suggestions! 

And... I initially just added kickstart file for EC2 with minimum change as I needed to figure out how imagefactory works for EC2.  Thank you for taking this task to actually add support for it! 😄

geertj:
@simaishi you're welcome :)

Regarding naming... Thinking about it some more, my preference would probably be for a short and obvious name, ideally the name of the compute service if that properly identifies the target. Using the compute service name instead of the cloud name has the benefit of allowing us to support other flavors of compute services for a cloud (maybe at some point we will need custom images for GKE and ECS). Azure would be the exception as the compute service there doesn't have a proper name (AFAIK it's just called "Compute").

I agree that `'<cloud>-<compute>'` is technically more correct, but it's longer and gets weird with e.g. `'gcp-gce'` and `'azure-compute'`.

Based on this would you be OK with renaming `'aws-ec2'` to `'ec2'`?

geertj:
redhat-imaging/imagefactory#385 just got merged. This PR can now be tested using the latest imagefactory master.

Fryguy:
'aws-ec2' to 'ec2' seems to be the least invasive change while achieving the best consistency. :+1:

jonnyfiveiq:
Set the image name to be 'ec2' for this release, though we need to start a migration to <vendor>-<service> such as Google-GCE or Google-GKE, VMware-vSphere or VMware-vCloud, as examples.

geertj:
Updated PR with `'aws-ec2'` renamed to `'aws'`.

geertj:
I added a WIP Ansible playbook for doing the upload to EC2. Some issues with the playbook still:

* It requires ansible/ansible#16138 which appears not yet available on centos 7.
* It doesn't yet do the actual AMI creation. Right now it stops at the EBS snapshot, from which the AMI should be just 1 API call removed. I'm having trouble understanding the `ec2_ami` module though for EBS backed AMIs.

geertj:
Removed the playbook for uploading images, as @simaishi was working on a different approach.

geertj:
Note that this requires https://github.com/redhat-imaging/imagefactory/issues/388.

simaishi:
@geertj Can you please update README for imagefactory SHA needed for EC2 image, and also add `python-boto` as a new RPM requirement for EC2 plugin?

simaishi:
Other than README needing an update, EC2 related changes look good. The only part I'm not sure is the ramfs driver change.  Some of the changes there aren't related to EC2 and I'd feel more comfortable someone else reviews that part of the change.  @Fryguy @abellotti ?

kickstarts/base.ks.erb
```diff
@@ -103,9 +103,19 @@ EOF
...
+   when "vsphere", "vagrant" %>
+drivers="mptbase mptscsih mptspi"
+<% when "azure", "hyperv" %>
+drivers="hv_storvsc hid_hyperv hv_netvsc hv_vmbus"
+<% when "ec2" %>
+drivers="xen-blkfront xen-netfront"
+<% else %>
+drivers=""
+<% end %>
+test -n "$drivers" && /sbin/dracut --force --add-drivers "$drivers" $ramfsfile $kversion
```
abellotti: why aren't we adding mptbase/mptscsih/mptspi any longer to non vsphere/vagrant ?

geertj: As far as I know that driver is a vSphere specific driver, and no other hypervisor that I am aware of uses it.

(It is used on bare metal. But that case is not relevant here.)

geertj: @abellotti we can re-enable it for all platforms if you prefer, but in my view it's slightly cleaner not to. The way I wrote the case statement makes it self documenting and we capture exactly which platforms need which drivers.

Also if those drivers are needed on another platform (pretty sure they aren't), the image won't boot at all. So it we would know very soon in any case.

When I said above vSphere, I meant vSphere and VirtualBox. I checked it and as far as I can see all other platforms use different drivers.

geertj:
@simaishi I made the changes as requested here https://github.com/ManageIQ/manageiq-appliance-build/pull/161#issuecomment-256622508

kickstarts/base.ks.erb
geertj: Updated PR so that we continue to install the LSI logic drivers on all platforms, so that we don't introduce big changes shortly before the Euwe release.

Will add submit a PR to master to remove them.

chessbyte:
Euwe Backport details:
```sh
$ git log -1
commit 65ad10f87824ebc4994ae83ef4f4c02cd19bcc3c
Author: Satoe Imaishi <simaishi@redhat.com>
Date:   Thu Nov 3 14:30:51 2016 -0400

    Merge pull request #161 from geertj/dev/ec2
    
    Add support for building EC2 images.
    (cherry picked from commit 41ec7f47127602d839db4cf980b5ea319b88c7f1)
```

PR #162, opened 2016-09-22
Don't hardcode branch name for docbuild

simaishi:
Pass in branch name (but default to 'master' if not passed), so this file doesn't need to be edited for each branch.

### DIFF
--- a/scripts/docbuild.rb
+++ b/scripts/docbuild.rb
@@ -7,7 +7,7 @@ $log = Logger.new(STDOUT)
 BUILD_BASE     = Pathname.new("/build")
 FILESHARE_DIR  = BUILD_BASE.join("fileshare")
 
-DOCS_BRANCH    = "master"
+DOCS_BRANCH    = ARGV[0] || "master"
 DOCS_REPO      = BUILD_BASE.join("manageiq_docs")
 DOCS_PACKAGE   = DOCS_REPO.join("_package/community", DOCS_BRANCH)
 DOCS_FILESHARE = FILESHARE_DIR.join(DOCS_BRANCH, "latest-docs")


simaishi:
@Fryguy please review.

scripts/docbuild.rb
```diff
@@ -7,7 +7,7 @@
 BUILD_BASE     = Pathname.new("/build")
 FILESHARE_DIR  = BUILD_BASE.join("fileshare")
 
-DOCS_BRANCH    = "master"
+DOCS_BRANCH    = ARGV[0].nil? ? "master" : ARGV[0]
```
abellotti: would prefer:  ARGV[0] || "master"

Thanks.

simaishi: Cleaner that way... done. Thanks!

Fryguy:
@simaishi You shouldn't need the branch at all.  asciibinder, by default, builds *all* of the branches.  You can build branch by branch if you want, so I'm not sure what you prefer.

Fryguy:
@simaishi I reviewed the current stuff, and I think it's better to remove the branches entirely.  asciibinder should not do a git checkout before a build because it needs things from the master branch (like some of the config files).

simaishi:
Wasn't aware it builds all branches by default. If that's the case, I think we can move up the `latest-docs` directory one level up (currently it's under 'master' directory), and put each branch doc to respective branch directory (which is related to https://github.com/ManageIQ/manageiq-appliance-build/pull/133).

simaishi:
@Fryguy @abellotti ascii_builder builds all branches in "local" git repo. So we need to make sure all needed branches exist locally, not just 'master'.  (Which is why I wasn't seeing 'euwe' being built after modifying distro_map yml)

So we can either hard-code needed branches somewhere, or build per branch from each nightly as per my original thought (by passing branch name).  Or maybe some other options?

Fryguy:
you should be able to `git fetch --all`, then run it.  That will get the branch locally, and the distro map yml is what drives the creation of the directories.

PR #167, opened 2016-09-30
Add optional post-build compression step

geertj:
There are a few current and future appliances where the output of imagefactory is a non-compressed format. This patch adds support for an optional post-build compression step to reduce the size of those images.

The supported formats are "gzip" for cases where the users are mostly Unix/Linux people, and "zip" for Windows folks.

This PR doesn't yet enable any compression. However I would suggest we enable in a future PR:
- ZIP compression for Azure (given the Windows audience). This will bring   down the appliance size of 3.3GB to a size similar to the other appliances  (currently roughtly 1GB for the nightly builds).
- ZIP compression for HyperV. Same situation as Azure.
- GZIP compression for EC2 (once EC2 PR has been enabled). The native image   format for EC2 is a sparse RAW file which will be 50GB in size. This is  not practical and must be compressed.

### DIFF
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -106,6 +106,7 @@ Dir.chdir(IMGFAC_DIR) do
   targets.sort.reverse.each do |target|
     imgfac_target = target.imagefactory_type
     ova_format    = target.ova_format
+    compression   = target.compression_type
     $log.info "Building for #{target}:"
 
     tdl_name = target.name == "azure" ? "base_azure.tdl" : "base.tdl"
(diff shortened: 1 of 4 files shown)


geertj:
@miq-bot add_label euwe/yes

jrafanie:
Seems reasonable but i'm not sure if this is desireable as you can't directly import them into the provider.

geertj:
@miq-bot remove_label euwe/yes

scripts/vmbuild.rb
```diff
@@ -146,7 +147,21 @@ def verify_run(output)
...
     file_name = "#{name}-#{target}-#{build_label}-#{timestamp}-#{manageiq_checkout.commit_sha}.#{target.file_extension}"
     destination = destination_directory.join(file_name)
-    $log.info `mv #{source} #{destination}`
+
+    case compression
+    when 'gzip'
+      $log.info `gzip -c #{source} > #{destination}.gz`
+      destination = destination.sub_ext(destination.extname + '.gz')
+    when 'zip'
+      $log.info `ln #{source} #{destination}`
```
simaishi: Hard-linking between `source` and `destination` will not always work. `source` is a temporary location and `destination` is where the final images are kept, which could be on different device (and that is the case for our build machine).

scripts/vmbuild.rb
```diff
@@ -146,7 +147,21 @@ def verify_run(output)
...
+      Dir.chdir destination_directory do
+        $log.info `zip #{destination.basename.sub_ext('.zip')} #{destination.basename}`
+      end
+      $log.info `rm #{destination}`
+      destination = destination.sub_ext('.zip')
+    else
+      $log.info `ln #{source} #{destination}`
```
simaishi: same here - can we use`mv` as that's what the existing code is doing anyway?

simaishi:
@geertj Because of [Azure VHD image issue](https://github.com/ManageIQ/manageiq-appliance-build/issues/212), I'd need to create Azure image a bit differently now and that will require the final image to be compressed. I'd like to use this new compress option if possible - would you mind updating the PR?

PR #169, opened 2016-10-04
added new node installation

chriskacerguis:
added new node installation method, to allow for Node 6.x

### DIFF
--- a/kickstarts/partials/post/node.ks.erb
+++ b/kickstarts/partials/post/node.ks.erb
@@ -1,0 +1,5 @@
+# Get the NodeJS RPM repo setup
+curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -
+
+# Install NodeJS 6.x (and npm)
+yum -y install nodejs
--- a/kickstarts/partials/post/node.ks.erb
+++ b/kickstarts/partials/post/node.ks.erb
@@ -1,5 +1,2 @@
-# Get the NodeJS RPM repo setup
-curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -
-
-# Install NodeJS 6.x (and npm)
-yum -y install nodejs
+# Install the NodeJS rpm (will include npm)
+rpm -i https://rpm.nodesource.com/pub_6.x/el/7/x86_64/nodejs-6.7.0-1nodesource.el7.centos.x86_64.rpm
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -76,7 +76,9 @@ export RAILS_USE_MEMORY_STORE="true"
 
 <%= render_partial "post/bundler" %>
 
-# needs to run *after* post/bundler
+<%= render_partial "post/node" %>
+
+# needs to run *after* post/bundler AND post/node
 <%= render_partial "post/ui_compile" %>
 
 # appliance_root="/opt/manageiq/manageiq-appliance" -- in post/source_setup partial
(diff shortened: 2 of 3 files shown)


kickstarts/partials/post/node.ks.erb
```diff
@@ -0,0 +1,5 @@
+# Get the NodeJS RPM repo setup
+curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -
```
carbonin: Is there a better way to do this? I don't particularly like feeding links directly to bash.

If this just installs a `.repo` file then we could probably add the repo url directly to https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/repos.ks.erb and then install the package in the package list.

kickstarts/partials/post/node.ks.erb
```diff
@@ -0,0 +1,5 @@
...
+
+# Install NodeJS 6.x (and npm)
+yum -y install nodejs
```
carbonin: Nit, but can we keep the newline at the end of the file?

Fryguy:
Related PR: https://github.com/ManageIQ/manageiq/pull/11683

kickstarts/partials/post/node.ks.erb
carbonin: I think typically the reason to avoid `curl <url> | bash -` is security as @Fryguy said.

This just also breaks from the pattern that we've established in general. If we could figure out what [this rpm](https://github.com/nodesource/distributions/blob/master/rpm/setup_6.x#L128) installs (my guess is a `.repo` file) then that would just solve it I think.

chriskacerguis: looks like the script, actually gets the rpm file itself, and not a .repo.  I'll update the PR

carbonin: Right, I suspect that `.rpm` file contains the repo file which we could just point to directly.

chriskacerguis: I'm checking on that.  FWIW - If someone can compromise that bash script, I'm sure they can compromise the source code to Node or the RPM.  So, yes blindly running a bash script maybe not the best, but just my $0.02

kickstarts/partials/post/node.ks.erb
```diff
@@ -0,0 +1,2 @@
+# Install the NodeJS rpm (will include npm)
+rpm -i https://rpm.nodesource.com/pub_6.x/el/7/x86_64/nodejs-6.7.0-1nodesource.el7.centos.x86_64.rpm
```
carbonin: Does this also install node? Previously we were running the script that installed this rpm, but then needed to also `yum -y install nodejs`.

carbonin: Alright, sounds good.

carbonin:
@simaishi Can you also review? Want to make sure this doesn't impact you as well.

simaishi:
EPEL now has v6.7 node RPM and we no longer need to get the RPM from a different repo.

geertj:
As a general rule, it would be great if we could install using a package manager (yum, gem, npm) as much as possible. As @Fryguy mentions, it allows us to reproduce builds with more certainty. It also gives us better provenance.

One specific one that I'm referring to is our source install of Ruby. Software collections now has a collection for 2.3 ("rh-ruby23"). This would mean we don't have to install and compile Ruby, and bundler. As a nice side effect, this would remove 10+ minutes from the build (my recent changes to PS4 allow you to see which post-install stuff takes most time).

chriskacerguis:
I would agree with @geertj if Ruby 2.3 is available via RPM (and has what we need), we should use that for sure.

chriskacerguis:
closing this.  Totally borked my local repo, so it's just easier to close and do a new one.

PR #171, opened 2016-10-04
Uptadte to v0.6.0

### LINKED ISSUE #170: Upgrade to ruby-install 0.6.0
ruby-install was updated to 0.6.0 a long while ago, and it supports the --update options, but we are still using 0.5.0 [here](https://github.com/ManageIQ/manageiq-appliance-build/blob/1f8fa5fbaf79d00cccec7c42034a2e0f54a1d8e9/kickstarts/partials/post/ruby_install.ks.erb#L1).  Upgrading should allow us to be more fluid with patch level versions of Ruby as indicated [here](https://github.com/ManageIQ/manageiq-appliance-build/blob/1f8fa5fbaf79d00cccec7c42034a2e0f54a1d8e9/kickstarts/partials/post/ruby_install.ks.erb#L12-L16)

Koltharius:
Fix #170


jrafanie:
Thanks for your contribution @Koltharius 🙇

PR #157, opened 2016-09-13
[DARGA] Reduce image size

geertj:
This is a backport of #149 to the Darga branch.

PR #172, opened 2016-10-05
[DARGA] Revert #157

Fryguy:
Something in #157 is causing the build to stall at the vm shutdown phase.  Reverting this PR so we can get darga-4 out the door.

cc @geertj


geertj:
Do we have log files? Should be somewhere in /root on the installed system (I believe /root/ks.cfg).

PR #177, opened 2016-10-18
Newest imagefactory has less dependencies.

geertj:
Work has been done in the imagefactory upstream not to have hard deps that
are only needed for the provider_image command (which isn't used by us).
Also the devinstall script has been updated.


Fryguy:
@simaishi Please review.

PR #179, opened 2016-10-25
Install unzip package

mikedahlgren:
Install the unzip package to allow for unzipping files and aid in installing agents on the appliance.


carbonin:
This package is present on the appliance currently.

PR #187, opened 2017-01-03
Install ruby-install from rpm rather than from source

carbonin:
Use the ruby-install rpm here https://copr-be.cloud.fedoraproject.org/results/postmodern/ruby-install/fedora-25-x86_64/00492646-ruby-install/

This way we don't have to build ruby-install from source every time we build an appliance.

### DIFF
--- a/kickstarts/partials/packages/includes.ks.erb
+++ b/kickstarts/partials/packages/includes.ks.erb
@@ -10,6 +10,7 @@ ipmitool
 # ruby dependencies auto-installed by ruby-install
 # prefer to have these dependencies explicitly required in the kickstart
 # https://github.com/postmodern/ruby-install/blob/8957e0d1a52d59b45cf48cef8f15a1d8ed921cd4/share/ruby-install/ruby/dependencies.txt
+ruby-install
 automake
 gcc
 gdbm-devel
--- a/kickstarts/partials/post/ruby_install.ks.erb
+++ b/kickstarts/partials/post/ruby_install.ks.erb
@@ -1,9 +1,3 @@
-curl -L https://github.com/postmodern/ruby-install/archive/v0.6.0.tar.gz | tar xvfzC - /tmp
-
-pushd /tmp/ruby-install*/
-make install
-popd
-
 # The ruby-install will result in about 350MiB of files in /usr/local/src.
 # These do not need to be in the image, so put them on a tmpfs.
 mkdir -p /usr/local/src
--- a/kickstarts/partials/post/ruby_install.ks.erb
+++ b/kickstarts/partials/post/ruby_install.ks.erb
@@ -7,7 +7,7 @@ mount -t tmpfs tmpfs /usr/local/src
 # Once the --update option is available, we could ask
 # ruby-install to "learn" what new rubies are available
 # See: https://github.com/postmodern/ruby-install/issues/175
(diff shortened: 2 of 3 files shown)


kickstarts/partials/packages/includes.ks.erb
```diff
@@ -10,7 +10,9 @@ ipmitool
 # ruby dependencies auto-installed by ruby-install
 # prefer to have these dependencies explicitly required in the kickstart
 # https://github.com/postmodern/ruby-install/blob/8957e0d1a52d59b45cf48cef8f15a1d8ed921cd4/share/ruby-install/ruby/dependencies.txt
```
carbonin: Not sure we need this as the ruby-install rpm should handle this right?

kickstarts/partials/post/ruby_install.ks.erb
```diff
@@ -1,23 +1,9 @@
...
-
-# The ruby-install will result in about 350MiB of files in /usr/local/src.
-# These do not need to be in the image, so put them on a tmpfs.
-mkdir -p /usr/local/src
-mount -t tmpfs tmpfs /usr/local/src
-
 # Note, we hardcode the ruby patch level for now.
 # Once the --update option is available, we could ask
 # ruby-install to "learn" what new rubies are available
 # See: https://github.com/postmodern/ruby-install/issues/175
```
carbonin: This issue was resolved, but I'm not sure how we wanted to use it here.

geertj: Does your change just install the "ruby-install" installer from RPM, or Ruby itself as well? Looking at the Copr I think it's the former, and the installer still downloads and installs ruby itself.

These lines are here so that the Ruby .tar.gz and the build files are not in the image. I was able to reduce the image by 350MB. Note that just deleting these files is not enough as the underlying storage will continue to be allocated.

geertj: Also in case this is just the installer - is there any way we could install Ruby from RPM from e.g. softwarecollections.org? My measurements indicated that this should reduce the build time of the application by about 7-10 minutes.

kickstarts/partials/packages/includes.ks.erb
carbonin: Ah I see, these are dependencies of ruby itself, not ruby-install

carbonin:
@geertj I added back the tmpfs lines you mentioned.

I'll move the conversation about where to install ruby from here so that it doesn't get clobbered by new commits.

The question raised by @geertj was:
>is there any way we could install Ruby from RPM from e.g. softwarecollections.org? My measurements indicated that this should reduce the build time of the application by about 7-10 minutes.

carbonin:
@geertj In general that's probably not a bad idea, but I think it's out of the scope of this change.

We would need to be sure that we properly enable the collection everywhere.
This change (ruby-install from rpm) is an easy win that can also be easily reverted if we decide to install ruby itself from a different source.

Removing WIP. I tested this out with a build.

kickstarts/partials/packages/includes.ks.erb
Fryguy: Yes these are ruby's dependencies, but even so, I wonder if the ruby-install rpm should preload those dependencies for us, because you can't build ruby without them (and the fact that it's an rpm implies this is a fedora/rhel platform target).

PR #188, opened 2017-01-03
Add the wget package

carbonin:
It is required for the build and was not previously required explicitly.

It must have been being included as a dependency at some point, but is longer present on appliances which is preventing us from downloading the copr repo files we need.

The `manageiq-ManageIQ-Fine-epel-7.repo` file is missing and I am seeing the following in `anaconda-post.log`
```
'/var/www/miq/lib' -> '/var/www/miq/vmdb/gems/pending'
+ [16:44:15] pushd /etc/yum.repos.d/
/etc/yum.repos.d /
+ [16:44:15] wget https://copr.fedorainfracloud.org/coprs/manageiq/ManageIQ-Fine/repo/epel-7/manageiq-ManageIQ-Fine-epel-7.repo
/tmp/ks-script-NH0n_X: line 104: wget: command not found
+ [16:44:15] popd
```
This also seems to affect euwe.

### DIFF
--- a/kickstarts/partials/packages/includes.ks.erb
+++ b/kickstarts/partials/packages/includes.ks.erb
@@ -49,6 +49,7 @@ scl-utils
 scl-utils-build                  # build requires
 socat                            # for proxying remote consoles
 telnet
+wget                             # build requires
 yum-utils
 zip                              # build requires
 


abellotti:
👍

simaishi:
Looks like `ipa-client` switched from using `wget` to `curl` in newer version.  Not sure if switching to `curl` in repos.ks.erb would be a better solution for us too?

carbonin:
@simaishi I was thinking the same, but do we want `wget` on the appliances regardless? I feel like it's something people might expect to be there. On a related note, is there a significant reason for using `curl` over `wget`?

simaishi:
After talking with Nick, we agreed adding `wget` back will be the desired behavior as something/someone might be relying on it..

simaishi:
Euwe backport details:

```
$ git log -1
commit 32c7f283746331572c8eb0d05eb0b7860245ba55
Author: Satoe Imaishi <simaishi@redhat.com>
Date:   Wed Jan 4 09:17:30 2017 -0500

    Merge pull request #188 from carbonin/add_wget_package
    
    Add the wget package
    (cherry picked from commit 5ab76d3d23e8182a2523bf9ff198f77c812e398e)
```

PR #193, opened 2017-02-23
For ovirt, create both ova and qcow2

simaishi:
Ovirt/RHV 4.1 supports qcow2, but not ova import. 3.x supports ova import, but not qcow2. So we need to create 2 images until all supported Ovirt/RHV versions can share 1 format.

Made it ovirt/qcow2 specific rather than adding 'generic' way to support multiple formats for a provider as:
- We're not planning to support 2 images for any other provider (we want to reduce # of images, not add more!)
- Even if we need to create 2 images for another provider in the future, depending on what needs to be built for which provider, process will be different

https://bugzilla.redhat.com/show_bug.cgi?id=1410020

### DIFF
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -129,7 +129,7 @@ Dir.chdir(IMGFAC_DIR) do
     $log.info "Running #{target} target_image #{imgfac_target} using parameters: #{params}"
 
     output = `./imagefactory --config #{IMGFAC_CONF} target_image #{params} --id #{uuid} #{imgfac_target}`
-    uuid   = verify_run(output)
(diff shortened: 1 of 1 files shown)


simaishi:
Euwe backport details:

```
$ git log -1
commit f6abbbe2f50e41becb21c61cc9df2fb4691d6379
Date:   Fri Feb 24 17:01:33 2017 -0500

    Merge pull request #193 from simaishi/ovirt_add_qcow_format
    
    For ovirt, create both ova and qcow2
    (cherry picked from commit c86c4727d19810975a8b29f6f765a16392e32cdb)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1427169
```

PR #194, opened 2017-02-27
Remove the symlink that used to point to gems/pending

carbonin:
The directory rooted at $app_root/vmdb/gems/pending has been moved
into a separate gem and is no longer present at that path.

We will not be updating the link location every time the gem
changes versions or we track a new git sha so I'm removing this.

### DIFF
--- a/kickstarts/partials/post/source_setup.ks.erb
+++ b/kickstarts/partials/post/source_setup.ks.erb
@@ -28,4 +28,3 @@ cp /etc/motd.manageiq /etc/motd
 
 # Legacy directory symlinks
 ln -vs $appliance_root $app_root/system
-ln -vs $app_root/vmdb/gems/pending $app_root/lib


carbonin:
https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/WriteVm/build_linux/init.rb#L13 is the only place I've found that seems to still rely on this.

Clearly it wouldn't have worked since the gem was split out, but worth mentioning here either way I think.

ping @roliveri

jrafanie:
@carbonin the [referenced link](https://github.com/ManageIQ/manageiq-appliance-build/pull/194#issuecomment-282748931) is from encrypted code days and includes a require to a now missing ruby class `MiqLoad` [here](https://github.com/ManageIQ/manageiq-gems-pending/blob/4b7fdb099632dc1b2a9aefc90db149c04f52a97c/lib/gems/pending/WriteVm/build_linux/init.rb#L8) so I highly doubt we need that file but I'll let @roliveri confirm

roliveri:
We don't need that file. Even back in the day, it was only a prototype.

Issue #127, opened 2016-04-29
Do not continue creating image when commands in %post section (e.g. bundle install) fails


simaishi:
Currently, we don't catch any error coming from commands ran in %post section of kickstart.  Some errors are fatal and appliance won't work at all, but we continue creating images anyway.

Would be nice to grab the anaconda-post.log, then stop the process, to avoid generating images that won't work.

mfeifer:
@simaishi what would you like to do with this old issue?

simaishi:
I've pinned the issue and assigned to myself. I'm still trying to figure out if this can be done, as this is really helpful if it can be done.

simaishi:
With rpm based appliance, we no longer need to get this done. What sometimes failed in the %post section is now all done during rpm build.

PR #206, opened 2017-04-04
bower - die when bower install fails

himdel:
Since the UI needs bower dependencies, it makes little sense to continue the build when bower has failed.

Similar to what we do on Travis - https://github.com/ManageIQ/manageiq-ui-classic/blob/master/tools/ci/setup_js_env.sh#L13-L18 .

@abellotti , @simaishi WDYT?

---

Before merging this, please note that I only think this should work, have not tested that `exit 1` in there stops the build. (I am sure that `rake update:bower` returns an exit code on most errors.)

### DIFF
--- a/kickstarts/partials/post/ui_compile.ks.erb
+++ b/kickstarts/partials/post/ui_compile.ks.erb
@@ -1,7 +1,8 @@
 npm install -g bower yarn
 
 pushd /var/www/miq/vmdb
-  rake update:bower
+  rake update:bower || exit 1
+
   RAILS_ENV=production rake evm:compile_assets
   rake evm:compile_sti_loader
 popd


simaishi:
I think I tried using `exit` before for https://github.com/ManageIQ/manageiq-appliance-build/issues/127 and remember that not working. I might have to add `--erroronfail`, but we do so much in %post and some might be non-fatal error we can ignore. I have to try it out...

himdel:
@simaishi Have you had a chance to try?

Should I close or rebase? :)

simaishi:
@himdel FYI, I did try with `exit 1` (as well as `--erroronfail`) but that wasn't enough to stop the build process... It stopped the %post script, but build itself continued anyway. Will have to dig into this more.

himdel:
Thanks, I hope we can drop bower in time for hammer so hopefull this shouldn't be an issue for much longer.

(That said, we should be able to catch yarn failing too.)

PR #207, opened 2017-04-06
Specified only production dependencies be installed

chalettu:
SUI fix for appliance build.  Ensured we only install production dependencies

### DIFF
--- a/kickstarts/partials/post/ui_compile.ks.erb
+++ b/kickstarts/partials/post/ui_compile.ks.erb
@@ -8,6 +8,6 @@ popd
 
 # Service UI
 pushd /opt/manageiq/manageiq-ui-service
-  yarn install
+  yarn install --production
   yarn run build
 popd


chalettu:
Dependent on a SUI PR https://github.com/ManageIQ/manageiq-ui-service/pull/648

Fryguy:
:+1:

Fryguy:
@simaishi Do we do bundle install in production mode for the manageiq repo?  I can't find it anywhere.  I see RAILS_ENV is set in /etc/default/evm in the manageiq-appliance repo, but that's it.

chriskacerguis:
👍

chriskacerguis:
Wondering if we should set the NODE_ENV on the appliance?

simaishi:
@Fryguy we do `bundle install` here: https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/bundler.ks.erb#L12

simaishi:
Confirmed working.  @chalettu Thanks!

PR #210, opened 2017-05-17
Generate the v2_key when we create the database

carbonin:
This is more similar to what the container build does and allows us to remove the v2_key generation from initialize_appliance.sh

### DIFF
--- a/kickstarts/partials/post/db_init.ks.erb
+++ b/kickstarts/partials/post/db_init.ks.erb
@@ -4,7 +4,7 @@ cat > /bin/appliance-initialize.sh <<EOF
 [[ -d /var/opt/rh/rh-postgresql95/lib/pgsql/data/base ]] && exit 0
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 echo "Initializing Appliance, please wait ..." > /dev/tty1
-appliance_console_cli --region 0 --internal --password smartvm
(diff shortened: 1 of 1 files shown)


kickstarts/partials/post/db_init.ks.erb
```diff
@@ -4,7 +4,7 @@ cat > /bin/appliance-initialize.sh <<EOF
 [[ -d /var/opt/rh/rh-postgresql95/lib/pgsql/data/base ]] && exit 0
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 echo "Initializing Appliance, please wait ..." > /dev/tty1
-appliance_console_cli --region 0 --internal --password smartvm
+appliance_console_cli --region 0 --internal --password smartvm --key
```
jrafanie: @carbonin We can't do it here right?  Won't all appliances have the same key since this is run during kickstart?

carbonin: This isn't run during the kickstart. We print the file here (I'm not sure why) and it runs on first boot.

This should really be moved out with the rest of our systemd unit files.

jrafanie: Oh right, I didn't see the heredoc until now.  I'm good with this then.  I saw "kickstarts/**/post" and was concerned.

PR #218, opened 2017-07-21
Delete temporary files created during build

simaishi:
Each nightly/release build uses about 180GB in STORAGE_DIR.  Multiple files are created during a build for each target, but only the final image is what we'd need to keep.

We clean up STORAGE_DIR daily using [clean_imagefactory_storage.sh](https://github.com/ManageIQ/manageiq-appliance-build/blob/master/bin/clean_imagefactory_storage.sh) so they're deleted anyway the next day. This change will clean up those files after each build rather than waiting for a day.

I know storage isn't that expensive now... but I have personally never used any of those temporary files within 24 hours after the build, and I didn't see a need to keep them around for a day.

### DIFF
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -177,7 +177,7 @@ Dir.chdir(IMGFAC_DIR) do
     end
 
     # The final image is moved out of STORAGE_DIR at this point, delete all other files created during build
-    temp_file_uuid.each { |uuid| FileUtils.rm(Dir.glob("#{STORAGE_DIR}/#{uuid}.*"), :verbose => true) }
+    temp_file_uuid.each { |file_uuid| FileUtils.rm_f(Dir.glob("#{STORAGE_DIR}/#{file_uuid}.*"), :verbose => true) }
   end
 
   passphrase_file = GPG_DIR.join("pass")
--- a/scripts/vmbuild.rb
+++ b/scripts/vmbuild.rb
@@ -124,6 +124,7 @@ Dir.chdir(IMGFAC_DIR) do
     output = `./imagefactory --config #{IMGFAC_CONF} base_image #{params} #{tdl_file}`
     uuid   = verify_run(output)
     $log.info "#{target} base_image complete, uuid: #{uuid}"
+    temp_file_uuid = [uuid]
 
     params = "--parameters #{target_file}"
     $log.info "Running #{target} target_image #{imgfac_target} using parameters: #{params}"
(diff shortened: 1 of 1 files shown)


simaishi:
@carbonin please review.  If we have a reason to keep those files, I'll close this PR and file a request to increase storage on our build machine so our nightly builds will go through 😄

scripts/vmbuild.rb
```diff
@@ -172,7 +175,11 @@ def verify_run(output)
         $log.info `scp #{destination} #{FILE_SERVER_ACCOUNT}@#{FILE_SERVER}:#{file_rdu_dir.join(file_name)}`
       end
     end
+
+    # The final image is moved out of STORAGE_DIR at this point, delete all other files created during build
+    temp_file_uuid.each { |uuid| FileUtils.rm(Dir.glob("#{STORAGE_DIR}/#{uuid}.*"), :verbose => true) }
```
carbonin: Does this need to be `FileUtils.rm_f`?

simaishi: It works with `FileUtils.rm` - probably because everything is ran and owned by root.

Fryguy:
>  Shadowing outer local variable - uuid.

Pick a different block variable name because it's colliding with the outer method's local variable.

scripts/vmbuild.rb
Fryguy: Do you need to know each specific uuid?  Can you just do `"#{STORAGE_DIR}/*")`

carbonin: Yeah, I think that would be better. This is what I'm seeing on my build machine even though I'm root ...

```plain
nick-brewery# ls
ansible_on_upstream.log
nick-brewery# whoami
root
nick-brewery# rm ansible_on_upstream.log 
rm: remove regular file ‘ansible_on_upstream.log’? y
```

simaishi: Updated to use `rm_f`.

I can't use "#{STORAGE_DIR}/*" because some files might be still in use.  For example, if 2 nightly builds are running at the same time (this is possible because we check if a build is running by `virsh list` but the 1st nightly build might be running a part of the process that doesn't initiate virsh process) when 2nd one started...  So I just need to delete what's used in that particular build.

simaishi:
Updated to use `rm_f`, also changed variable name.

simaishi:
Fine backport details:

```
$ git log -1
commit 339256ebd0e9a7cd64e6215450de80c3ffac4fcc
Date:   Wed Jul 26 09:15:14 2017 -0400

    Merge pull request #218 from simaishi/delete_temp_files
    
    Delete temporary files created during build
    (cherry picked from commit b3045b9b3642ba8e8b681224614144bc9cb2d962)
```

PR #220, opened 2017-08-02
Use update:ui task, instead of update:bower

simaishi:
Mirrors the change in https://github.com/ManageIQ/manageiq/pull/15578


himdel:
@miq-bot add_label gaprindashvili/yes, transformation

simaishi:
This is already in gaprindashvii branch.

PR #228, opened 2017-09-27
Add qpid-proton-c libs

gberginc:
qpid-proton-c is required by the Nuage provider to support AMQP eventing (ActiveMQ). The cyrus-sasl packages are required to support the authentication mode Nuage network provider uses in their validated deployments.


gberginc:
@simaishi based on your offline comments, I have removed the `swig` dependency.

PR #229, opened 2017-10-03
Upgrade to Ruby 2.4

chrisarcand:
**To be merged on Wednesday, October 4th**

Updates our stuff to use Ruby 2.4.2

Part of https://github.com/ManageIQ/manageiq/issues/14446

I don't know if I'm missing anything else, please let me know if I am.

### DIFF
--- a/kickstarts/partials/post/ruby_install.ks.erb
+++ b/kickstarts/partials/post/ruby_install.ks.erb
@@ -7,13 +7,13 @@ mount -t tmpfs tmpfs /usr/local/src
 # Once the --update option is available, we could ask
 # ruby-install to "learn" what new rubies are available
 # See: https://github.com/postmodern/ruby-install/issues/175
-ruby-install ruby 2.3.1 -- --disable-install-doc --enable-shared
+ruby-install ruby 2.4.2 -- --disable-install-doc --enable-shared
 
 # Free up tmpfs memory.
 umount /usr/local/src
 
 # Add the ruby binaries path to the PATH so we can find bundle and friends:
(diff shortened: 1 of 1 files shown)


chrisarcand:
cc/ @jrafanie

Fryguy:
Since we are so close to the release, we may want to wait until we branch for gaprindashvili, and put this on master only.  @chessbyte Thoughts?

Fryguy:
IMO The Oct 4 date is more about removing 2.2 support, than moving all the things to 2.4.  We can still keep the appliances at 2.3 for support sanity, but drop Ruby 2.2 support on Oct 4.

chrisarcand:
? I asked about this from the beginning and the consent was that I wouldn't have to wait for the G-release branch as this is long overdue. I'm confused about the upstream/downstream split, but will ask offline. 👍

jrafanie:
Fyi, this is the reason we can't upgrade to 2.4 in rpmland:
https://bugzilla.redhat.com/show_bug.cgi?id=1438736

There isn't a rpm for 2.4.2.  2.4.0 has serious bugs so we can't use it.

Fryguy:
>  I asked about this from the beginning and the consent was that I wouldn't have to wait for the G-release branch as this is long overdue

Just to clarify, we can support Ruby 2.3 and 2.4 Oct 4 just fine.  I just think we should stick with Ruby 2.3 for the G build.  So *this* PR should wait, but the rest can move forward, IMO.

chrisarcand:
Clarified all sorts of things and ultimately, from this PR's view, 2.3.1 should remain for now as @Fryguy says. 🏃

PR #230, opened 2017-10-11
Use a master copr repo on the master branch

carbonin:
From now on, we can fork the master repo into a branch name repo
at the same time that we create the release branch in this repo.

This allows us to change package versions without affecting previous
releases more easily.


carbonin:
Should be merged with https://github.com/ManageIQ/manageiq/pull/16176

carbonin:
Waiting to test and build newly released pglogical 2.1.0

carbonin:
Built and tested pglogical 2.1.0.

This is good to go :smile:

PR #231, opened 2017-10-16
Add libguestfs tools for OpenStack OpenSCAP

aufi:
Adding libguestfs-tools package to be able mount OpenStack snapshots
in way suitable to perform OpenSCAP scan.

### DIFF
--- a/kickstarts/partials/packages/includes.ks.erb
+++ b/kickstarts/partials/packages/includes.ks.erb
@@ -124,3 +124,6 @@ cockpit-ssh
 cyrus-sasl
 cyrus-sasl-plain
 qpid-proton-c-devel
+
+# For OpenStack snapshots mount for OpenSCAP scan
+libguestfs-tools


aufi:
@simaishi May I ask for review? Thanks!

carbonin:
Is there a related app PR which will use this new package?

Also, will we need to add this package to the container build [here](https://github.com/ManageIQ/manageiq-pods/blob/master/images/miq-app/Dockerfile#L44)?

aufi:
There is no final PR which makes it working yet, but the tracking can be done via issue  https://github.com/ManageIQ/manageiq-providers-openstack/issues/125

I will submit manageiq-pods PR soon, thanks for mentioning it!

Fryguy:
@roliveri Please review.

roliveri:
Would The libguestfs VM image actually run in the MIQ appliance, or would this be used to interface with an external instance?

aufi:
@roliveri The libguestfs should mount an image on the MIQ appliance.

roliveri:
@aufi I have a number of concerns about this.

Unless things have changed a great deal, as I recall, libguestfs has a large number of dependencies, and is fairly heavyweight in resource consumption. Unless this is intended to only be used on specific appliance instances, dedicated to OpenScap, it could have a negative impact on the other workers running on the appliance.

In addition, I'm not sure how this would fit into the new MIQ architecture. You may want to consider implementing this as a containerized service that is deployed external to the appliance. Then, it may naturally fit into the new architecture.

I would suggest presenting an overview of the entire design to a larger MIQ audience. There are people who know more about the appliance and re-architecture than I do. They may be able to help identify possible issues early on.

aufi:
@roliveri Hi, I agree that libguestfs package is not single purpose lightweight solution.

From OpenStack point of view, ```qemu-image``` package would be enough - would it change something for you? Thanks.

roliveri:
@aufi 
Sorry I haven't replied sooner - my laptop died, and I'm still catching up after setting up the replacement.

I'm not that familiar with `qemu-image` so I can't say for sure. However, I would still have concerns surrounding how this fits into the new MIQ architecture.

I still suggest presenting an overview of the design to a larger MIQ audience so this can be assessed.

aufi:
@roliveri OK, thanks for info. I quickly discussed it with @Ladas yesterday, we'll try get larger audience.

Ladas:
@aufi could you create a RFC in here https://github.com/ManageIQ/manageiq-design/issues and describe in detail how the OpenSCAP scan should work? That is what @roliveri is asking for :-)

aufi:
@Ladas ah, will do, thanks!

Issue #219, opened 2017-07-26
4.5G root partition too small for vmware appliance


tjyang:
Hi 

* This current miq instance with master branch without upgrade on miq01 machine.
```
[root@miq01 ~]# df -lh
Filesystem                              Size  Used Avail Use% Mounted on
/dev/mapper/vg_system-lv_os             4.5G  3.8G  718M  85% /
<snipped>

[root@miq01 ~]# ls /opt/
manageiq  rh  rubies
[root@miq01 ~]#

```
* On another instance miq02, I  am out of root diskspace easily by  upgrade command.
```
vmdb && systemctl  stop evmserverd && git pull && bin/update && reboot
```
*  Please increase / size, if I am not doing anything wrong.

tjyang:
Hi @simaishi ,

Can the disk layout  have  extra 10-20G  unassigned space  ?
So one can assign the free space to vg_system or vg_data volume groups when needed.

```
[root@miq01 log]# pvs
  PV         VG        Fmt  Attr PSize  PFree
  /dev/sda2  vg_system lvm2 a--  26.00g 140.00m
  /dev/sda5  vg_data   lvm2 a--  13.50g  20.00m

[root@miq01 ~]# lvs
  LV               VG        Attr       LSize   Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  lv_pg            vg_data   -wi-ao----  13.48g                                 
  lv_home          vg_system -wi-ao----   1.00g                                 
  lv_os            vg_system -wi-ao----   4.50g                                 
  lv_swap          vg_system -wi-ao----   5.86g                                 
  lv_tmp           vg_system -wi-ao----   1.00g                                 
  lv_var           vg_system -wi-ao----   2.00g                                 
  lv_var_log       vg_system -wi-ao----  11.00g                                 
  lv_var_log_audit vg_system -wi-ao---- 512.00m                                 
[root@miq01 ~]#

```

simaishi:
@tjyang We actually have a need to increase /var size as well, so let me talk to a few people and figure out what we need to do. I like your idea of leaving some 'open' space 😄

tjyang:
@simaishi, we have this practice at work of leaving open space, so admin can extend and resize lv on the fly.  I can easily add /dev/sdb vmdk file but that will make this appliance with two vmdk files  hard/impossible to do v2v(VMware to OpenStack) later.

PR #232, opened 2017-10-17
Increase filesystem sizes

carbonin:
This PR increases the `/` filesystem from 4.5GiB to 10GiB and the `/var` filesystem from 2GiB to 12GiB

The root filesystem change is needed so that upgrades can be performed by using `bin/update` as described in #219. In this case the new gems being installed in `/opt/rubies/ruby-2.3.1/lib/ruby/gems` were filling up the space on the root filesystem.  It should be noted that after upgrading the appliance, we can run `bundle clean --force` to remove old versions of updated gems and reclaim some space.

The `/var` filesystem change is needed so that we can use embedded ansible. Both the rpm and docker versions require more space. The rpm version has an explicit check which requires at least 10GiB of free space on the `/var` partition. We currently ignore that check [here](https://github.com/ManageIQ/manageiq/blob/5f3a0ce3a23f3f59d250269c06ac00c86582f006/lib/embedded_ansible.rb#L143), but it isn't checking without reason, so we would like to stop doing that.
Additionally the container version of embedded ansible (AWX) which is being developed in https://github.com/ManageIQ/manageiq/pull/16205 will need more space in `/var` for the docker daemon.

Adding WIP until my test build finishes.

### DIFF
--- a/config/base.tdl
+++ b/config/base.tdl
@@ -11,6 +11,6 @@
   </os>
   <description>CentOS73 TDL meant to be used with custom kickstart for manageiq.</description>
   <disk>
-    <size>56G</size>
+    <size>66G</size>
   </disk>
 </template>
--- a/config/base.tdl
+++ b/config/base.tdl
@@ -11,6 +11,6 @@
   </os>
   <description>CentOS73 TDL meant to be used with custom kickstart for manageiq.</description>
   <disk>
-    <size>50G</size>
+    <size>56G</size>
   </disk>
 </template>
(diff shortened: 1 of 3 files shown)


carbonin:
This is what the appliance filesystems look like now.

screenshot from 2017-10-17 18-07-00

Removing WIP.

simaishi:

```
$ git log -1
commit 0614eda1576dddad4a1bc1fbabfa8119af42066b
Author: Gregg Tanzillo <gtanzill@redhat.com>
Date:   Fri Nov 3 11:33:33 2017 -0400

    Merge pull request #232 from carbonin/increase_filesystem_sizes
    
    Increase filesystem sizes
    (cherry picked from commit 5da3b7b0e52964f048c5bde74529e61fb16ba79a)
```

PR #236, opened 2017-10-23
add BUILD file to miq checkout

durandom:
this helps QE and developers to identify the correct build date of upstream appliances

related PR https://github.com/ManageIQ/manageiq/pull/16196


durandom:
@simaishi @Fryguy I have no env to test this, but it works from the cmd line :)
Maybe you can try this out?

PR #244, opened 2017-11-10
Update to CentOS 7.4

simaishi:
Closes https://github.com/ManageIQ/manageiq-appliance/issues/137

I will put a follow up PR to use just 1 tdl file for all images. It's probably not worth keeping 2 files, just to save 5GB disk size. (As you can see in the diff, I forgot to update azure one when we went from 7.2 to 7.3!)

The new ISO is already on the build machine.

### DIFF
--- a/config/base.tdl
+++ b/config/base.tdl
@@ -1,15 +1,15 @@
 <template>
-  <name>CentOS-7.3</name>
+  <name>CentOS-7.4</name>
   <os>
     <rootpw>ozrootpw</rootpw>
     <name>CentOS-7</name>
-    <version>3</version>
+    <version>4</version>
     <arch>x86_64</arch>
     <install type="iso">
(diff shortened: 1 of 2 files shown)


carbonin:
Just assuming `gaprindashvili/no` is that right @simaishi ?

simaishi:
I'd say `gaprindashvili/yes` to keep upstream/downstream in sync.

simaishi:

```
$ git log -1
commit 7032387a11f0bb11226de5d1c02465558da6bfb9
Date:   Fri Nov 10 08:54:36 2017 -0500

    Merge pull request #244 from simaishi/centos74
    
    Update to CentOS 7.4
    (cherry picked from commit 143cd7ab9f1e96254fe1e1d4556fc64adcbdbc01)
```

PR #259, opened 2018-01-11
Use noop io scheduler by default

carbonin:
KB article from VMware which recommends the noop scheduler: https://kb.vmware.com/s/article/2011861

Instructions on how to set the default io scheduler from Red Hat: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/performance_tuning_guide/sect-red_hat_enterprise_linux-performance_tuning_guide-storage_and_file_systems-configuration_tools#sect-Red_Hat_Enterprise_Linux-Performance_Tuning_Guide-Configuration_tools-Setting_the_default_IO_scheduler

Wondering if we think we should only change this for VMware images or for all of them?

https://bugzilla.redhat.com/show_bug.cgi?id=1531088

### DIFF
--- a/kickstarts/partials/main/bootloader.ks.erb
+++ b/kickstarts/partials/main/bootloader.ks.erb
@@ -1,9 +1,9 @@
 <% if @target == "ec2" || @target == "openstack" %>
-bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=tty0 console=ttyS0,115200n8"
+bootloader --location=mbr --driveorder=vda --append="elevator=noop crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=tty0 console=ttyS0,115200n8"
 <% elsif @target == "azure" %>
-bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
+bootloader --location=mbr --driveorder=vda --append="elevator=noop crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
 <% elsif @target == "gce" %>
-bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=ttyS0,38400n8"
+bootloader --location=mbr --driveorder=vda --append="elevator=noop crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=ttyS0,38400n8"
 <% else %>
-bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0"
+bootloader --location=mbr --driveorder=vda --append="elevator=noop crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0"
 <% end %>


jrafanie:
> Wondering if we think we should only change this for VMware images or for all of them?

Great question @carbonin. Who would know?

> KB article from VMware which recommends the noop scheduler: https://kb.vmware.com/s/article/2011861

Maybe I'm missing it but it looks like they're recommending either NOOP or Deadline

> Testing has shown that NOOP or Deadline performs better for virtualized Linux guests

carbonin:
> Who would know?

I'm not sure anyone would. It's more of a risk calculation than a question I expect hard data on.

Do we think this could cause something bad on other platforms? Or is the idea of doing less "smart" things with respect to lower level operations a good idea regardless of the virtualization platform?

> it looks like they're recommending either NOOP or Deadline

Indeed, but because we were using deadline before and the BZ lists noop as causing an "incredible increase in performance" my comment was more "VMware says noop is *also* recommended for good performance" rather than "VMware clearly recommends noop over deadline"

rvalente:
@jrafanie @carbonin so the risk of using a blanket [noop] scheduler is if the customer has a raw device map instead of a virtual machine. Essentially what we're trying to accomplish by using [noop] is allow the hypervisor to handler all optimizations to the physical devices, instead of doing it twice, once in the guest and once on the host.

tldr; there is risk, only if the customer is using SR-IOV or RDM.

rvalente:
So I have done more digging into the ManageIQ appliance. Setting this option manually is mutually exclusive to running `tuned` and the appliances ship with `tuned` installed and running be default.

```bash
$ rpm -qa | grep tuned
tuned-2.8.0-5.el7_4.2.noarch

$ systemctl status tuned
● tuned.service - Dynamic System Tuning Daemon
   Loaded: loaded (/usr/lib/systemd/system/tuned.service; enabled; vendor preset: enabled)
   Active: active (running) since Thu 2018-03-01 07:58:38 EST; 1 day 8h ago
 Main PID: 1373 (tuned)
   CGroup: /system.slice/tuned.service
           └─1373 /usr/bin/python -Es /usr/sbin/tuned -l -P
```

Here are the list of profiles available, the active one (as expected) is `virtual-guest`.

```bash
$ tuned-adm list
Available profiles:
- balanced                    - General non-specialized tuned profile
- desktop                     - Optimize for the desktop use-case
- latency-performance         - Optimize for deterministic performance at the cost of increased power consumption
- network-latency             - Optimize for deterministic performance at the cost of increased power consumption, focused on low latency network performance
- network-throughput          - Optimize for streaming network throughput, generally only necessary on older CPUs or 40G+ networks
- powersave                   - Optimize for low power consumption
- throughput-performance      - Broadly applicable tuning that provides excellent performance across a variety of common server workloads
- virtual-guest               - Optimize for running inside a virtual guest
- virtual-host                - Optimize for running KVM guests
Current active profile: virtual-guest
```

If the `noop` scheduler were to be set then we would want to remove the `tuned` package. I am not against this as it would open up the option to make more purposeful changes to improve performance in the future.

## PostgreSQL Performance and I/O Schedulers
- [linux-io-scheduler-tuning](https://blog.codeship.com/linux-io-scheduler-tuning/)
- [postgresql-io-schedulers-cfq-noop-deadline](https://blog.pgaddict.com/posts/postgresql-io-schedulers-cfq-noop-deadline)

rvalente:
Pulled this from the `tuned` instance running on my ManageIQ appliance:

`
virtual-guest/tuned.conf `

```ini
#
# tuned configuration
#

[main]
summary=Optimize for running inside a virtual guest
include=throughput-performance

[sysctl]
# If a workload mostly uses anonymous memory and it hits this limit, the entire
# working set is buffered for I/O, and any more write buffering would require
# swapping, so it's time to throttle writes until I/O can catch up.  Workloads
# that mostly use file mappings may be able to use even higher values.
#
# The generator of dirty data starts writeback at this percentage (system default
# is 20%)
vm.dirty_ratio = 30

# Filesystem I/O is usually much more efficient than swapping, so try to keep
# swapping low.  It's usually safe to go even lower than this on systems with
# server-grade storage.
vm.swappiness = 30
```

`throughput-performance/tuned.conf`

```ini
#
# tuned configuration
#

[main]
summary=Broadly applicable tuning that provides excellent performance across a variety of common server workloads

[cpu]
governor=performance
energy_perf_bias=performance
min_perf_pct=100

[disk]
readahead=>4096

[sysctl]
# ktune sysctl settings for rhel6 servers, maximizing i/o throughput
#
# Minimal preemption granularity for CPU-bound tasks:
# (default: 1 msec#  (1 + ilog(ncpus)), units: nanoseconds)
kernel.sched_min_granularity_ns = 10000000

# SCHED_OTHER wake-up granularity.
# (default: 1 msec#  (1 + ilog(ncpus)), units: nanoseconds)
#
# This option delays the preemption effects of decoupled workloads
# and reduces their over-scheduling. Synchronous workloads will still
# have immediate wakeup/sleep latencies.
kernel.sched_wakeup_granularity_ns = 15000000

# If a workload mostly uses anonymous memory and it hits this limit, the entire
# working set is buffered for I/O, and any more write buffering would require
# swapping, so it's time to throttle writes until I/O can catch up.  Workloads
# that mostly use file mappings may be able to use even higher values.
#
# The generator of dirty data starts writeback at this percentage (system default
# is 20%)
vm.dirty_ratio = 40

# Start background writeback (via writeback threads) at this percentage (system
# default is 10%)
vm.dirty_background_ratio = 10

# PID allocation wrap value.  When the kernel's next PID value
# reaches this value, it wraps back to a minimum PID value.
# PIDs of value pid_max or larger are not allocated.
#
# A suggested value for pid_max is 1024 * <# of cpu cores/threads in system>
# e.g., a box with 32 cpus, the default of 32768 is reasonable, for 64 cpus,
# 65536, for 4096 cpus, 4194304 (which is the upper limit possible).
#kernel.pid_max = 65536

# The swappiness parameter controls the tendency of the kernel to move
# processes out of physical memory and onto the swap disk.
# 0 tells the kernel to avoid swapping processes out of physical memory
# for as long as possible
# 100 tells the kernel to aggressively swap processes out of physical memory
# and move them to swap cache
vm.swappiness=10
```

PR #264, opened 2018-04-25
Clear yarn cache

simaishi:
Clearing yarn cache reduces the uncompressed image size by ~800MB

Uncompressed 'master' images (hyperv, azure, etc.) are now so big that it exceeds rackspace size limit and not being uploaded there since February... We should/can probably clean up more, but starting with this easy one.

### DIFF
--- a/kickstarts/partials/post/ui_compile.ks.erb
+++ b/kickstarts/partials/post/ui_compile.ks.erb
@@ -11,3 +11,6 @@ pushd /opt/manageiq/manageiq-ui-service
   yarn install --production
   yarn run build
 popd
+
+# Clean cache, will be populated again when yarn runs next time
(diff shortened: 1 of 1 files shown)


carbonin:
Good find! Can we take some of the cleanup we do for the docker builds and apply that here?
Should be some more easy wins in that comparison.

simaishi:

```
$ git log -1
commit 75b5c22791e926efb69ba5cdfb7c74dc52a87faa
Date:   Wed Apr 25 09:59:46 2018 -0400

    Merge pull request #264 from simaishi/yarn_clean
    
    Clear yarn cache
    (cherry picked from commit 82ae7bb8af8b3ed742b717471445356f26d9dcac)
```

PR #270, opened 2018-06-15
Install ansible and python-boto for running ansible-playbook

simaishi:
For https://github.com/ManageIQ/manageiq/pull/17564. Both packages are available in EPEL.

cc @Ladas

### DIFF
--- a/kickstarts/partials/packages/includes.ks.erb
+++ b/kickstarts/partials/packages/includes.ks.erb
@@ -126,3 +126,7 @@ cockpit-ssh
 cyrus-sasl
 cyrus-sasl-plain
 qpid-proton-c-devel
+
+# For ansible-playbook
+ansible
+python-boto


Ladas:
Looks good :+1: 

Will we do more PRs for that packages needed by the other providers?

simaishi:
@Ladas I'm not sure what other packages you're talking about... I'm adding the two you mentioned in https://github.com/ManageIQ/manageiq/pull/17564, and not aware of anything else needed!

Ladas:
@simaishi right, we will need more, based on what other providers will need, lets revive the mail conversation about it

carbonin:
@Ladas @simaishi do we still need this if we're moving to ansible-runner?

Ladas:
@carbonin I think we can close this

- ansible-runner beats the ansible rpm
- python-boto is needed for AWS playbooks, but we will not be merging those

carbonin:
Okay, closing for now until we decide what will be needed for runner.

Provider dependencies can probably be handled separately as well.

PR #273, opened 2018-07-24
Compile static assets in production for Node as well

skateman:
Hopefully this fixes at least some of styling issues on the appliances. Credit goes to @hyperkid123, @himdel could you please verify if this helps?

screenshot from 2018-07-24 09-37-35


Hyperkid123:
we might also need to define a plugin in webpack to make this work properly

skateman:
Not needed

himdel:
Fixed by https://github.com/ManageIQ/manageiq-ui-classic/pull/4341

PR #277, opened 2018-08-22
Add dracut-config-generic for EC2 ENA support

simaishi:
Also sorted the list in the 2nd commit...

https://bugzilla.redhat.com/show_bug.cgi?id=1574029

cc @jrafanie


simaishi:
@jrafanie Yes, safe to backport, adding the label.

Issue #26, opened 2015-08-21
Appliance Properties shows "Red Hat Enterprise Linux 6 64 bit"


abellotti:
Downloading/Importing CentOS 7.1 built appliance ova to Fusion, the General properties of an appliance shows "Red Hat Enterprise Linux 6 64-bit" should be the "... 7 64-bit" string.  not sure where the property comes from, not in our config files, could be set in Imagefactory/Oz land.

JPrause:
Spoke with Alberto and the issue as described still exists.

Fryguy:
@simaishi Bump 🐢   This issue as opened in a slightly different way in ManageIQ/manageiq-appliance#121

simaishi:
ImageFactory doesn't expose the OS setting and it's hard-coded as "rhel6_64Guest".

https://github.com/redhat-imaging/imagefactory/blob/master/imagefactory_plugins/ovfcommon/ovfcommon.py#L465-L469

The only way to 'fix' this currently is to overwrite the value in Imagefactory source that's on our build machine...

simaishi:
I see an open issue in imagefactory: https://github.com/redhat-imaging/imagefactory/issues/401

PR #278, opened 2018-08-23
Update vSphere ova

simaishi:
A few changes for vSphere appliance image.

For https://bugzilla.redhat.com/show_bug.cgi?id=1429653:
- Changed hardware version to vmx-10 (was: vmx-7/vmx-8)
- Changed network to VmxNet3 (was: E1000)
- Added paravirtual drivers

I didn't make the default driver to be paravirtual in OVA (default: LSI Logic) as that will make the image not to work on Fusion 10. Since the drivers are now included in the image, anyone wanting to use paravirtual can just change the setting once imported to vSphere.

For https://bugzilla.redhat.com/show_bug.cgi?id=1614006:
- Set OS type to `centos64Guest` (requires imagefactory [c78579f](https://github.com/redhat-imaging/imagefactory/commit/c78579ff3e7e3ed5fa79fab912968381d486f608) or later). OS will show as `CentOS 4/5/6/7 (64-bit)` in vSphere and `CentOS 64-bit` in Fusion. This fixes https://github.com/ManageIQ/manageiq-appliance-build/issues/26 cc @abellotti 

I will update imagefactory on build machine once this is merged.

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -102,12 +102,15 @@ EOF
 # Let's rebuild the ramfs with with base scsi drivers we need
 kversion=$(rpm -q kernel --qf '%{version}-%{release}.%{arch}\n')
 ramfsfile="/boot/initramfs-$kversion.img"
-drivers="mptbase mptscsih mptspi"
 <% case @target
    when "azure", "hyperv" %>
-drivers="$drivers hv_storvsc hid_hyperv hv_netvsc hv_vmbus"
+drivers="mptbase mptscsih mptspi hv_storvsc hid_hyperv hv_netvsc hv_vmbus"
 <% when "ec2" %>
-drivers="$drivers xen-blkfront xen-netfront"
+drivers="mptbase mptscsih mptspi xen-blkfront xen-netfront"
+<% when "vsphere" %>
+drivers="mptspi vmw_pvscsi"
+<% else %>
+drivers="mptbase mptscsih mptspi"
 <% end %>
 /sbin/dracut --force --add-drivers "$drivers" $ramfsfile $kversion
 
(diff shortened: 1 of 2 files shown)


kickstarts/base.ks.erb
```diff
@@ -102,13 +102,17 @@ EOF
 # Let's rebuild the ramfs with with base scsi drivers we need
 kversion=$(rpm -q kernel --qf '%{version}-%{release}.%{arch}\n')
 ramfsfile="/boot/initramfs-$kversion.img"
+<% if @target == "vsphere" %>
```
carbonin: Can we add this to the existing `case @target` statement on line 109 rather than a separate `if ... else ... end`?

simaishi: https://github.com/ManageIQ/manageiq-appliance-build/pull/278/files#diff-97b85e7883f17153493a362c11fb3fe3L105 apply to all but vsphere, so I moved vsphere up outside of `case`, but I can change if the other way reads better.

carbonin: The multiple levels of indentation are hard to read here. I think I would rather have all the drivers listed out by target and just remove the initial assignment of `drivers="vmw_pvscsi mptspi"`. So something like:

```erb
<% case @target
   when "azure", "hyperv" %>
drivers="mptspi mptbase mptscsih hv_storvsc hid_hyperv hv_netvsc hv_vmbus"
<% when "ec2" %>
drivers="mptspi mptbase mptscsih xen-blkfront xen-netfront"
<% when "vsphere" %>
drivers="mptspi vmw_pvscsi"
<% end %>
/sbin/dracut --force --add-drivers "$drivers" $ramfsfile $kversion
```

What do you think?

simaishi: Moved `vsphere` to `case` statement.

PR #280, opened 2018-09-21
Update appliances to ruby 2.4.4

jrafanie:
@simaishi I just noticed this is still 2.3.1... figured we wanted 2.4.4 on master/hammer, right?  Are there other places that need to be bumped to match what's available downstream?


simaishi:
@jrafanie yes, we should have `2.4.4` for master... thanks!

We should update for podified images too...  cc @carbonin

PR #281, opened 2018-09-21
Add RPMs needed for Ansible playbooks and roles for oVirt

simaishi:
Add python-ovirt-engine-sdk4
https://bugzilla.redhat.com/show_bug.cgi?id=1610768

Add ovirt-ansible-roles
https://bugzilla.redhat.com/show_bug.cgi?id=1610798

cc @mwperina @Fryguy

### DIFF
--- a/kickstarts/partials/post/ovirt_ansible.ks.erb
+++ b/kickstarts/partials/post/ovirt_ansible.ks.erb
@@ -1,0 +1,5 @@
+# Add oVirt repository
+yum install -y https://resources.ovirt.org/pub/yum-repo/ovirt-release42.rpm
+
+# For oVirt Ansible playbooks and roles
+yum install -y python-ovirt-engine-sdk4 ovirt-ansible-roles
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -71,6 +71,8 @@ export RAILS_USE_MEMORY_STORE="true"
 
 <%= render_partial "post/repos" %>
 
+<%= render_partial "post/ovirt_ansible" %>
+
 <%= render_partial "post/ruby_install" %>
 
 <%= render_partial "post/bundler" %>


kickstarts/partials/post/ovirt_ansible.ks.erb
```diff
@@ -0,0 +1,5 @@
+# Add oVirt repository
+yum install -y https://resources.ovirt.org/pub/yum-repo/ovirt-release42.rpm
```
mwperina: Just please don't forget that oVirt is upstream only repository, which is fine for ManageIQ. But for CFME we should be using downstream builds from RHV erratas.

simaishi: yup, downstream will be taken care of differently!

carbonin: `https://resources.ovirt.org/pub/yum-repo` is called `yum-repo` ... and it isn't a repo? :laughing: 

@mwperina is this really the case?

Also, I can't seem to download the rpm ... or get to https://resources.ovirt.org/ at all ... is this website even live?

```
$ wget https://resources.ovirt.org/pub/yum-repo/ovirt-release42.rpm
--2018-09-24 11:28:52--  https://resources.ovirt.org/pub/yum-repo/ovirt-release42.rpm
Resolving resources.ovirt.org (resources.ovirt.org)... 66.187.230.28
Connecting to resources.ovirt.org (resources.ovirt.org)|66.187.230.28|:443... connected.
GnuTLS: Error in the pull function.
Unable to establish SSL connection.
```
:point_up: took a bit over a minute

```
sudo yum install https://resources.ovirt.org/pub/yum-repo/ovirt-release42.org
[sudo] password for ncarboni: 
Last metadata expiration check: 2:14:20 ago on Mon 24 Sep 2018 09:16:55 AM EDT.
[MIRROR] ovirt-release42.org: Curl error (28): Timeout was reached for https://resources.ovirt.org/pub/yum-repo/ovirt-release42.org [Operation timed out after 30001 milliseconds with 0 out of 0 bytes received] 
[FAILED] ovirt-release42.org: Curl error (28): Timeout was reached for https://resources.ovirt.org/pub/yum-repo/ovirt-release42.org [Operation timed out after 30001 milliseconds with 0 out of 0 bytes received] 
Curl error (28): Timeout was reached for https://resources.ovirt.org/pub/yum-repo/ovirt-release42.org [Operation timed out after 30001 milliseconds with 0 out of 0 bytes received]
```

As it stands it looks like this might just break the build or in the best case scenario just wouldn't work ...

simaishi: Maybe a temporary network issue? I did a full build with this change on Friday, and everything worked ok...

carbonin: Could be. I was trying to pull the RPM to get the contents of the repo it provides to see if we could do something more clever here, but I still can't get to it.

mwperina: The RPM is official supported way how to setup on a computer the repositories required for oVirt installation. Installing ovirt-release42 is easiest way how to get installed latest version of python SDK and ansible roles.

There was an issues with HTTP server, I've ping our administrators and the issue was resolved.

carbonin:
Okay, I think we'll go with this for now. The repo structure looks non-trivial so this is probably the best we can do.

As @bdunne mentioned in the BZ we originally planned to create something within the plugin mechanism which would allow plugins to dictate their own list of dependent rpms, but since we haven't gotten around to designing such a feature (let alone implementing it) this will have to do.

PR #283, opened 2018-09-25
Changes for hammer branch release.

Fryguy:
@simaish @carbonin @bdunne Please review.

I believe this will require us to create a new copr repo, but I don't know how to do that.


carbonin:
Created the new copr repo here https://copr.fedorainfracloud.org/coprs/manageiq/ManageIQ-Hammer/

simaishi:
Can we also change `config/ova.json` and `kickstarts/partials/post/repos.ks.erb`

PR #292, opened 2018-10-09
Add vspk python module for Nuage

simaishi:
https://bugzilla.redhat.com/show_bug.cgi?id=1630801

Bambou (which is also listed in the BZ) is a dependency of vspk.

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -73,6 +73,8 @@ export RAILS_USE_MEMORY_STORE="true"
 
 <%= render_partial "post/ovirt_ansible" %>
 
+<%= render_partial "post/python_modules" %>
+
 <%= render_partial "post/ruby_install" %>
 
 <%= render_partial "post/bundler" %>
--- a/kickstarts/partials/post/python_modules.ks.erb
+++ b/kickstarts/partials/post/python_modules.ks.erb
@@ -1,0 +1,4 @@
+yum install -y python-pip
+
+# For Nuage
+pip install vspk==5.3.2


carbonin:
Is this going to change our upgrade procedure? Will we need to start upgrading the pip packages in some way? Also, are there any version constraints? I know the BZ says the latest version, but that seems awfully optimistic... Are they sure that no new version will break their integration?

carbonin:
cc @jrafanie @Fryguy thoughts on this?

jrafanie:
> Is this going to change our upgrade procedure? Will we need to start upgrading the pip packages in some way? Also, are there any version constraints? I know the BZ says the latest version, but that seems awfully optimistic... Are they sure that no new version will break their integration?

These are all good questions.  Can we get @Loicavenel to comment here?  While, we can certainly install the latest pip and vspk, it's going to the "latest" packaged versions, which is usually super old.  It would be great to know what minimum versions we need and if there are any consumers of it to help verify it still works on upgrades.  "For nuage" doesn't really tell us much.

simaishi:
vspk is coming straight from pypi, so it will be the latest available at the time of appliance build.

I think the future plan is to use requirements.txt, and we'll be able to specify version requirement there. So the concern for now will really be if we might take packages that's newer than what we want to use.

cc @miha-plesko

jrafanie:
> vspk is coming straight from pypi, so it will be the latest available at the time of appliance build.
> 
> I think the future plan is to use requirements.txt, and we'll be able to specify version requirement there. So the concern for now will really be if we might take packages that's newer than what we want to use.

Right, my bad. Only python-pip will be from a package.  The means each build of manageiq may get different versions of vspk depending on how frequently it's changed.

simaishi:
> The means each build of manageiq may get different versions of vspk depending on how frequently it's changed.

I think that itself isn't a problem as that's how gem (as well as rpm) works too. The only difference here is that we can't prevent breaking version coming to the build as we don't have requirements.txt yet..

@miha-plesko, I think your input will be needed here... if you'd like version constraints to be set now, we can do that, please let me know what it should be. Or we can leave it open and assume breaking change in vspk (if any) will come with code update...

miha-plesko:
@simaishi lets lock it to `vspk==5.3.2`, please. We wrote "latest" to the BZ long time ago when we didn't know when the integration will be complete. Well, it is complete enough now that we can lock the version.

simaishi:
Updated to lock to `5.3.2`

simaishi:
Hammer backport details:

```
$ git log -1
commit ddec56e5c4b759aa210ebffa7db833ab01e65412
Date:   Fri Oct 12 17:10:35 2018 -0400

    Merge pull request #292 from simaishi/add_python_vspk
    
    Add vspk python module for Nuage
    
    (cherry picked from commit 146dae92db0153117f36bdb771e862eec2976e7d)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1630801
```

PR #290, opened 2018-10-08
Set commit_sha just once

simaishi:
We're getting new git sha every time commit_sha is called. This causes appliance file names to be different for the same nightly build run if a commit happens when build is running:

manageiq-openstack-hammer-201810050800-1310fd6601.qc2
manageiq-ovirt-hammer-201810050800-28a765dd38.ova

This only affects file names and contents are ok as all kickstart files are generated at the same time.

Diff will be easier with [?w=1](https://github.com/ManageIQ/manageiq-appliance-build/pull/290/files?w=1)

PR #295, opened 2018-10-24
[Changelog] Add Sprint 97 and Hammer Beta1

JPrause:
[skip ci]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - Add post compression support [(#288)](https://github.com/ManageIQ/manageiq-appliance-build/pull/288)
 - Compress Azure, HyperV and EC2 images [(#289)](https://github.com/ManageIQ/manageiq-appliance-build/pull/289)
 - Move from apache module mod_auth_kerb to mod_auth_gssapi [(#282)](https://github.com/ManageIQ/manageiq-appliance-build/pull/282)
+- Add dracut-config-generic for EC2 ENA support [(#277)](https://github.com/ManageIQ/manageiq-appliance-build/pull/277)
 
 ### Fixed
 - Update vSphere ova [(#278)](https://github.com/ManageIQ/manageiq-appliance-build/pull/278)
 - Increase filesystem sizes [(#232)](https://github.com/ManageIQ/manageiq-appliance-build/pull/232)
 - Set commit_sha just once [(#290)](https://github.com/ManageIQ/manageiq-appliance-build/pull/290)
+- Update to node 8 [(#276)](https://github.com/ManageIQ/manageiq-appliance-build/pull/276)
 
 ## Gaprindashvili-4 - Released 2018-07-16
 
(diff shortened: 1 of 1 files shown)


JPrause:
@miq-bot add_label documentation

JPrause:
@miq-bot add_label hammer/no

CHANGELOG.md
```diff
@@ -5,31 +5,33 @@ All notable changes to this project will be documented in this file.
...
+- Move from apache module mod_auth_kerb to mod_auth_gssapi [(#282)](https://github.com/ManageIQ/manageiq-appliance-build/pull/282)
 
 ### Fixed
 - Update vSphere ova [(#278)](https://github.com/ManageIQ/manageiq-appliance-build/pull/278)
+- Increase filesystem sizes [(#232)](https://github.com/ManageIQ/manageiq-appliance-build/pull/232)
+- Set commit_sha just once [(#290)](https://github.com/ManageIQ/manageiq-appliance-build/pull/290)
 
-## Unreleased as of Sprint 91 ending 2018-07-30
+### Removed
+- Remove non-existing "server-policy" yum group [(#268)](https://github.com/ManageIQ/manageiq-appliance-build/pull/268)
```
carbonin: This line should be removed as this fix first appeared in Gaprindashvili-4.

carbonin: How did it even get here? This wasn't in an "unreleased" section before this PR ...

simaishi:
It looks some PRs are missing... e.g. https://github.com/ManageIQ/manageiq-appliance-build/pull/277

carbonin:
@simaishi that should probably be in a Gaprindashvili release though right? I would prefer "fixes" to the changelog be made in separate PRs.

simaishi:
No, that's in Hammer Beta1, so it got missed in the PR that went to Hammer branch too.

JPrause:
What probably happened with PR277, was the Milestone label was not added. As such, it may not have shown up in the list of PRs for that particular sprint.

simaishi:
@JPrause I think Sprint 92 and/or Sprint 93 got missed.. Looks like there are a few other PRs missing from the same time period. Regardless, they can be added now as needed!

JPrause:
@simaishi there are certain PRs (based on labels) that not included in the changelog. This is intentional, and has been in place since before I've been doing the changelogs. So I'd say, let's move on and get these all merged.

simaishi:
@JPrause I really hope we'll change the process to use a label for changelog. Otherwise, the reviewers can't tell why certain PR is included in the changelog and not, and questions like this will come up again. It won't matter if we're going to change the process, but if not, can you at least document how label affects changelogs?

Please do add https://github.com/ManageIQ/manageiq-appliance-build/pull/277 and https://github.com/ManageIQ/manageiq-appliance-build/pull/276 (I'm not sure why this isn't listed as this is labeled as `bug` and has Sprint 92). Those are more important changes (which affect appliances) than PRs like https://github.com/ManageIQ/manageiq-appliance-build/pull/290!

JPrause:
@simaishi sure and no problem. Even (and hopefully we do) after we implement a label system, the reviewer/merger can always request a particular PR be added, if they forgot to add the changelog label at merge time.

I'll add the PRs you identified to this PR.

In the case of PR277, there was no milestone set. It never showed up on my list because of that omission. PR276 was excluded since it has a label of dependencies. (I'll add the list of labels that are excluded to the process doc)

PR #305, opened 2019-01-08
Update .travis.yml to current ruby versions

carbonin:


carbonin:
@jrafanie any ideas for this travis failure?

Looks like the rake version is real old. Why would that be?

jrafanie:
@carbonin Try bundle update in the scripts directory and check in the Gemfile.lock.  It's not been updated in 3 years.  

https://github.com/ManageIQ/manageiq-appliance-build/commit/9b66dc82e68828ab591ec1ab295e2d3474f29f99#diff-ac1b7edefdf0a7fc3999ab078c92279b

Or better yet, maybe we can even remove the Gemfile.lock?

carbonin:
Oh wow. I didn't realize that we checked that in here. @simaishi I think you're the main user of this, when do we typically change that?

simaishi:
I don't really use that file. I think that's there just for spec? Then I'm not sure if we really need to keep the lock file committed...

jrafanie:
> I don't really use that file. I think that's there just for spec? Then I'm not sure if we really need to keep the lock file committed...

Agreed.  Delete it.  The intention was to know what was used to run the build process but that was when I was running builds. ✂️  🔥 🚽

PR #307, opened 2019-03-05
Temporarily downgrade qpid proton c library

simaishi:
EPEL has updated qpid-proton-c to 0.26.0 recently. But this version doesn't seem to be compatible with gem version we lock to (0.22.0) and "bundle install" is currently failing for both `master` and `hammer` branches on appliances:

```
Installing qpid_proton 0.22.0 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
......
cproton.c: In function ‘_wrap_pn_selectables’:
cproton.c:16896:3: error: unknown type name ‘pn_selectables_t’
......
An error occurred while installing qpid_proton (0.22.0), and Bundler cannot continue.
```

To get back to working state quickly, temporarily downgrading the c library package to 0.19.0 that's available in ManageIQ copr repo and is known to work.

@Fryguy @carbonin please reivew

@gberginc @miha-plesko Just the other day, we discussed how updating the qpid_proton version could be a pain, but now we have found out NOT updating is causing a problem too.  I think downgrading a rpm package isn't best practice and would like to avoid if possible.  Can we look into updating the gem version?

### DIFF
--- a/kickstarts/partials/post/bundler.ks.erb
+++ b/kickstarts/partials/post/bundler.ks.erb
@@ -10,6 +10,9 @@ gem install bundler -v ">=1.8.4"
 
 ln -s ${APPLIANCE_SOURCE_DIRECTORY}/manageiq-appliance-dependencies.rb /var/www/miq/vmdb/bundler.d/manageiq-appliance-dependencies.rb
 
+# TEMPORARY FIX
+yum -y downgrade qpid-proton-c-0.19.0-1.el7.centos qpid-proton-c-devel-0.19.0-1.el7.centos
+
 pushd /var/www/miq/vmdb
   bundle install --with qpid_proton
   bundle clean --force


miha-plesko:
@simaishi any chance you build an appliance with 0.26.0 for us? Looking at changelog I assume nothing broke the interface so hopefully it'll just work. We have all the functionality covered with integration tests so if you give us the appliance, we can quite easily test.

simaishi:
Hammer backport details:

```
$ git log -1
commit fcf4d47859d69a6b952f4aa8587c91b297e56563
Date:   Tue Mar 5 10:01:04 2019 -0500

    Merge pull request #307 from simaishi/downgrade_qpid_proton
    
    Temporarily downgrade qpid proton c library
    
    (cherry picked from commit dcd4044671ef15c9bb712eff74d26b9472ce4554)
```

PR #315, opened 2019-04-05
[WIP] Enable NetworkManager

carbonin:
Running `ifup eth0` without it complains that it isn't running

Seems that now we're at the point @kbrock was suggesting in
https://github.com/ManageIQ/manageiq-appliance-build/pull/66
and we need NetworkManager enabled to get an ip address

@simaishi this should fix your issue with the downstream appliances not having an IP address, but I'm not sure what it will do to the upstream ones .... It should be fine .. right?

### DIFF
--- a/kickstarts/partials/post/systemd.ks.erb
+++ b/kickstarts/partials/post/systemd.ks.erb
@@ -16,5 +16,3 @@ ln -sf /dev/null /etc/systemd/system/ctrl-alt-del.target
 
 # Create the journal directory to enable persistant logging
 mkdir /var/log/journal
-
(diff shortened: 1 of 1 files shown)


carbonin:
Yeah, I think we'll hold off on this until we can take a closer look into the implications for the upstream appliance. For now @simaishi is handling the change separately.

simaishi:
Re-opening, as this will be probably needed when we move to CentOS 8.

simaishi:
This change is included in the CentOS 8 update PR: https://github.com/ManageIQ/manageiq-appliance-build/pull/381

PR #318, opened 2019-05-14
[WIP] Bind mount the log dir instead of using a symlink

agrare:
Using a symlink to the log directory leads to a dirty git tree, if we
add a log/.keep and also touch /var/log/manageiq/.keep we can bind mount
the /var/log/manageiq directory without dirtying the git status and also
remove the issues with not having a log/ directory in any of the repos
necessitating a bin/setup script to get tests to run.

TODO:

- [ ] Add the bind mount to /etc/fstab so the log dir is mounted on boot

### DIFF
--- a/kickstarts/partials/post/source_setup.ks.erb
+++ b/kickstarts/partials/post/source_setup.ks.erb
@@ -20,8 +20,8 @@ ln -vs $vmdb_root $manageiq_root
 
 # Replace the log directory with the larger logvol
 mkdir -p /var/log/manageiq
-rm -rf $vmdb_root/log
-ln -vs /var/log/manageiq $vmdb_root/log
+touch /var/log/manageiq/.keep
+mount --bind /var/log/manageiq $vmdb_root/log
 
 <%= git_checkout(@sui_checkout, "$sui_root") %>
 


agrare:
Re: https://github.com/ManageIQ/manageiq/issues/18639#issuecomment-492295555

cc @Fryguy @simaishi

carbonin:
Will this cause an issue when we try to mount a new volume on the appliance?

That's the only other place I feel like we might need to consider.
Specifically here https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/logfile_configuration.rb#L96

agrare:
Hm in that case there isn't already a var_log logical volume?

In the upstream appliance build the log volume is precreated https://github.com/ManageIQ/manageiq-appliance-build/blob/cd0e4736ca33c797f8cf8550f4ca35b9407b04fc/kickstarts/partials/main/disk_layout.ks.erb#L12

I'm thinking the way to deal with this generally is to create a systemd mount unit that evmserverd would depend on so that it always gets mounted before starts, that way appliance console can setup the log disk after the system image is created.  That way we don't have to modify `/etc/fstab` at runtime but I'm open to other ideas (including just doing what we do now).

carbonin:
Yeah, if you use the option I think you end up with an unused volume upstream. I don't actually recall how it works downstream ... @simaishi do we also ship with a volume for the logs there or do we expect people to provide something for that?

As for fstab the configuration in the console updates that at runtime here https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/logical_volume_management.rb#L81

simaishi:
> Yeah, if you use the option I think you end up with an unused volume upstream. I don't actually recall how it works downstream ... @simaishi do we also ship with a volume for the logs there or do we expect people to provide something for that?

Just like upstream, we ship downstream with a volume.

agrare:
Should appliance-console be adding the disk to the var_log_vg then vg_resize && lv_resize then?

carbonin:
Probably, but I think that's a separate thing, right @agrare? While it would definitely be an improvement I don't think it will help with the git status thing you're trying to solve, right?

agrare:
Oh correct it wouldn't fix the `git status` issue, it would simplify things in that we wouldn't need to unmount the bind, mount the new log disk, then remount the bind if someone added a disk with appliance_console though

carbonin:
Yeah, but I'm not even sure what's going on with that option as it is....
How would the symlink interact with us trying to mount the new volume on `/var/www/miq/vmdb/log`?

Seems like the whole thing needs to be rethought either way.

agrare:
:scream: I think it would mount it on top of what log is symlinked to (/var/log/manageiq)

agrare:
Yup that's what it does, so you'd have something like
```
/dev/mapper/vg_system-lv_var_log on /var/log
/dev/mapper/vg_miq_logs-lv_miq_logs on /var/log/manageiq through /var/www/miq/vmdb/log
```

carbonin:
Well at least we're not throwing away the rest of the space in the filesystem created for `/var/log` but still ... ew.

agrare:
Yeah I don't think that was intentional, also it would mount over any existing logs that were there from before creating that disk so they wouldn't be visible nor would they ever be rotated/purged

kickstarts/partials/post/source_setup.ks.erb
```diff
@@ -20,8 +20,8 @@ ln -vs $vmdb_root $manageiq_root
 
 # Replace the log directory with the larger logvol
 mkdir -p /var/log/manageiq
-rm -rf $vmdb_root/log
-ln -vs /var/log/manageiq $vmdb_root/log
+touch /var/log/manageiq/.keep
```
kbrock: I like the mount much better. can we drop the `touch` though?

kbrock: huh - looks like `.keep` is back in the repo. this works for me

PR #320, opened 2019-05-22
Update to node 10

simaishi:
Updating node to 10.x to match downstream appliance.

cc @himdel


himdel:
Thanks!, LGTM :)

(node **12** will depend on https://github.com/ManageIQ/manageiq-ui-classic/issues/5589, but this one is fine :))

PR #322, opened 2019-05-31
Add cockpit packages and changes to allow it to run by default

carbonin:
This should replace a good amount of the system management
we do in the appliance console

Requires https://github.com/ManageIQ/manageiq-appliance/pull/239 for cockpit to work

### DIFF
--- a/kickstarts/partials/post/firewalld.ks.erb
+++ b/kickstarts/partials/post/firewalld.ks.erb
@@ -16,3 +16,4 @@ firewall-offline-cmd --zone=manageiq --add-service=https
 firewall-offline-cmd --zone=manageiq --add-service=ssh
 firewall-offline-cmd --zone=manageiq --add-service=postgresql
 firewall-offline-cmd --zone=manageiq --add-service=dhcpv6-client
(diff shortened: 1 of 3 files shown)


kickstarts/partials/post/firewalld.ks.erb
```diff
@@ -16,3 +16,4 @@ firewall-offline-cmd --zone=manageiq --add-service=https
 firewall-offline-cmd --zone=manageiq --add-service=ssh
 firewall-offline-cmd --zone=manageiq --add-service=postgresql
 firewall-offline-cmd --zone=manageiq --add-service=dhcpv6-client
+firewall-offline-cme --zone=manageiq --add-service=cockpit
```
simaishi: @carbonin `firewall-offline-cme` typo...

carbonin: :man_facepalming:

carbonin: https://github.com/ManageIQ/manageiq-appliance-build/pull/323

PR #324, opened 2019-06-21
pip install django for ansible credential upgrades

carbonin:
This is needed for a python script we use to migrate credentials
which are encrypted using the awx encryption algorithm which relies
on some django utilities

If we change that migration to vendor the django utilities or
rewrite the whole thing in ruby before we release it then this
can be removed.


carbonin:
Don't need this since @NickLaMuro worked out how to do the decryption in ruby.

PR #325, opened 2019-07-08
Do not use unrecommended mount option nobarrier

jvlcek:
Starting with RHEL6 the mount option nobarrier was no longer recommended.
Starting with RHEL8 the mount option nobarrier was no longer supported. The mount
command fails on RHEL8 if the nobarrier option is provided.

This PR removes the unrecommended nobarrier mount option.

This will be more important once we can build upstream against a CentOS version 8.


jvlcek:
@miq-bot add_label hammer/no

jvlcek:
@miq-bot assign @carbonin

PR #340, opened 2019-08-07
[WIP] Add ansible-tower-venv-ansible rpm for extra modules

carbonin:
This provides the modules we need to support the roles behind
the additional credential types.

https://bugzilla.redhat.com/show_bug.cgi?id=1734129


carbonin:
@simaishi mentioned to me that this is actually the wrong GPG key for this package so the install will fail.

I think she's looking into if we should even use this package or if we should just use a different key.

carbonin:
Based on what @simaishi found with the rpm from releases.ansible.com I think we're going to go the direction of creating our own venv with the modules we need.

So I think I'm gonna just close this one and open a new PR when I have that sorted out.

PR #341, opened 2019-08-08
Install python modules required for embedded ansible credentials

carbonin:
We install these into a new venv so that we don't conflict with
appliance python packages.

This list is taken directly from
https://raw.githubusercontent.com/ansible/awx/c7bb0f10e10476e10df8a7e7b26732eb60958ca1/requirements/requirements_ansible.txt
which is the result of resolving the dependencies from
https://github.com/ansible/awx/blob/c7bb0f10e10476e10df8a7e7b26732eb60958ca1/requirements/requirements_ansible.in

I feel like taking the resolved file first is safer since pip doesn't
do true dependency resolution (pypa/pip#988)

Replacement for https://github.com/ManageIQ/manageiq-appliance-build/pull/340

Looking for input on places to put the list of packages. I felt like people wouldn't go looking for this list in `manageiq-appliance`, but it needs to live on disk somewhere ...

I could put it in manageiq core, but this seemed like a better place to start since it's really a build-time concern.

https://bugzilla.redhat.com/show_bug.cgi?id=1734129

### DIFF
--- a/kickstarts/partials/post/python_modules.ks.erb
+++ b/kickstarts/partials/post/python_modules.ks.erb
@@ -5,3 +5,130 @@ pip install vspk==5.3.2
 
 # For Lenovo
 pip install pylxca==2.1.1
+
+# For embedded ansible
+pip install virtualenv
+
+mkdir -p /var/lib/manageiq/
+pushd /var/lib/manageiq
+  virtualenv venv
+  source venv/bin/activate
+
+  cat > requirements.txt << EOF
+adal==1.2.1
+apache-libcloud==2.5.0
+appdirs==1.4.3
+applicationinsights==0.11.1
+argcomplete==1.9.4
+asn1crypto==0.24.0
+azure-cli-core==2.0.35
+azure-cli-nspkg==3.0.2
+azure-common==1.1.11
+azure-graphrbac==0.40.0
+azure-keyvault==1.0.0a1
+azure-mgmt-authorization==0.51.1
+azure-mgmt-batch==5.0.1
+azure-mgmt-cdn==3.0.0
+azure-mgmt-compute==4.4.0
+azure-mgmt-containerinstance==1.4.0
+azure-mgmt-containerregistry==2.0.0
+azure-mgmt-containerservice==4.4.0
+azure-mgmt-cosmosdb==0.5.2
+azure-mgmt-devtestlabs==3.0.0
+azure-mgmt-dns==2.1.0
+azure-mgmt-hdinsight==0.1.0
+azure-mgmt-keyvault==1.1.0
+azure-mgmt-loganalytics==0.2.0
+azure-mgmt-marketplaceordering==0.1.0
+azure-mgmt-monitor==0.5.2
+azure-mgmt-network==2.3.0
+azure-mgmt-nspkg==2.0.0
+azure-mgmt-rdbms==1.4.1
+azure-mgmt-redis==5.0.0
+azure-mgmt-resource==2.1.0
+azure-mgmt-servicebus==0.5.3
+azure-mgmt-sql==0.10.0
+azure-mgmt-storage==3.1.0
+azure-mgmt-trafficmanager==0.50.0
+azure-mgmt-web==0.41.0
+azure-nspkg==2.0.0
+azure-storage==0.35.1
+backports.ssl-match-hostname==3.5.0.1
+bcrypt==3.1.4
+boto==2.47.0
+boto3==1.6.2
+botocore==1.9.3
+cachetools==3.0.0
+certifi==2018.1.18
+cffi==1.11.5
+chardet==3.0.4
+colorama==0.3.9
+cryptography==2.6.1
+decorator==4.2.1
+deprecation==2.0
+docutils==0.14
+dogpile.cache==0.6.5
+google-auth==1.6.2
+humanfriendly==4.8
+idna==2.6
+ipaddress==1.0.19
+iso8601==0.1.12
+isodate==0.6.0
+Jinja2==2.10.1
+jmespath==0.9.3
+jsonpatch==1.21
+jsonpointer==2.0
+keystoneauth1==3.11.2
+knack==0.3.3
+lxml==4.1.1
+MarkupSafe==1.1.1
+monotonic==1.4
+msrest==0.6.1
+msrestazure==0.5.0
+munch==2.2.0
+ncclient==0.6.3
+netaddr==0.7.19
+netifaces==0.10.6
+ntlm-auth==1.0.6
+oauthlib==2.0.6
+openstacksdk==0.23.0
+os-service-types==1.2.0
+ovirt-engine-sdk-python==4.2.4
+packaging==17.1
+paramiko==2.4.2
+pbr==3.1.1
+pexpect==4.6.0
+psutil==5.4.3
+ptyprocess==0.5.2
+pyasn1==0.4.2
+pyasn1-modules==0.2.3
+pycparser==2.18
+Pygments==2.2.0
+PyJWT==1.6.0
+pykerberos==1.2.1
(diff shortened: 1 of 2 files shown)


carbonin:
So I made all the changes needed to get `pip install` to succeed without warnings or errors on the appliance while keeping as many of the modules as I could.

Unfortunately, now, a significant number are different from what is listed in awx and also from what is shipped in the tower rpm. Plus we added a few additional rpms.

Not sure what option we want to go with here. I can always revert this PR to its original state though if we think that's better. @simaishi @NickLaMuro @Fryguy ?

NickLaMuro:
@carbonin Yeah, I don't know.  Personally, I would rather we be as close to downstream as possible, or there eventually will be someone who tries fixing/replicating a bug on upstream and then turns out it is an issue with these packages and fixing it in one spot doesn't solve it in the other.

I am going to have to come back to this, but this how I started evaluating it:

- Took a look at what changed from the first pass (using `awx` versus the Tower rpm):
  
  https://github.com/carbonin/manageiq-appliance-build/compare/9bb4516...f367e74
  
  Which looks pretty bad, but when looking at
  
- The changes that got the "straight from tower" list to our new our current list (starting one commit up):
  
  https://github.com/carbonin/manageiq-appliance-build/compare/cab28a77...f367e74
  
  There is mostly removals from the list.

So I want to take a finer look at the about, but have to work on a few of my own TODO items first.

carbonin:
Yeah, I came to the same conclusion @NickLaMuro I think we're closer to the downstream set with what we have here even though it required a bunch of workarounds vs the upstream set.

I'm going to un-wip this and if we need to follow up on it we can do that after. Better to just get cloud modules working more quickly.

kickstarts/partials/post/python_modules.ks.erb
```diff
@@ -5,3 +5,167 @@ pip install vspk==5.3.2
...
+
+mkdir -p /var/lib/manageiq/
+pushd /var/lib/manageiq
+  virtualenv venv
+  source venv/bin/activate
+
+  cat > requirements.txt << EOF
+adal==1.2.1
+ansible==2.8.3
+ansible-runner==1.3.4
```
NickLaMuro: Just caught this while doing another pass.  We probably don't wants these in here, right?

carbonin: They are there in the downstream RPM so I left them in the interest of consistency.

NickLaMuro: Just hoping that these won't conflict with what is installed with the `ansible-runner` package in `kickstarts/partials/packages/includes.ks.erb`

carbonin: It won't. This will install binstubs in the venv directory. So we won't get this one unless we were to activate the venv before running `ansible-runner` (which we wont).

In fact we will never actually activate the venv. We are only passing the internal directory as a part of the `PYTHONPATH` so the stubs for these modules never actually get on our `PATH` when we're executing runner.

To illustrate this is from the appliance I've been doing the testing for this PR on:
```
[root@localhost venv]# pwd
/var/lib/manageiq/venv
[root@localhost venv]# which ansible-runner
/usr/bin/ansible-runner
[root@localhost venv]# source bin/activate
(venv) [root@localhost venv]# which ansible-runner
/var/lib/manageiq/venv/bin/ansible-runner
(venv) [root@localhost venv]# deactivate
[root@localhost venv]# which ansible-runner
/usr/bin/ansible-runner
```

simaishi:
I'll leave it up to @carbonin and @NickLaMuro for what should be included in requirements.txt, the rest looks good to me.

kickstarts/partials/post/python_modules.ks.erb
carbonin: Plus if this is a problem here then it will be a problem in CF as well so it's probably worth dealing with it in both places if it is an issue. As you pointed out in https://github.com/ManageIQ/manageiq-appliance-build/pull/341#issuecomment-519687490 :trollface:

kickstarts/partials/post/python_modules.ks.erb
```diff
@@ -5,3 +5,167 @@ pip install vspk==5.3.2
...
+
+# For embedded ansible
+pip install virtualenv
+
+mkdir -p /var/lib/manageiq/
+pushd /var/lib/manageiq
+  virtualenv venv
+  source venv/bin/activate
+
+  cat > requirements.txt << EOF
```
Fryguy: Minor, but would this be better as a straight file in COPY?

carbonin: I considered that, but thought that it might be more natural to keep what is essentially a packages list in this repo rather than in `manageiq-appliance`.

carbonin:
@simaishi brought up that `ansible` and `ansible-runner` are not actually present in the tower rpm. I was using the list from `pip freeze` to create this, but that also seems to be listing modules installed at the system level (not just the ones in the venv). I can re-work this to try to just install the modules included in the venv in the rpm (not using `pip freeze`) ... thoughts?

carbonin:
`pip freeze -l` is a thing. So now I have a new an updated list of modules that are actually installed using the tower venv rpm. Updating this PR to use that list.

Fryguy:
well, this got a lot simpler :joy:

kickstarts/partials/post/python_modules.ks.erb
Fryguy: I think this is resolved now that we don't have them in the list after using `pip freeze -l`

NickLaMuro: > Plus if this is a problem here then it will be a problem in CF as well so it's probably worth dealing with it in both places if it is an issue. As you pointed out in #341 (comment) :trollface:

So... you are saying you believe me now :trollface:

kickstarts/partials/post/python_modules.ks.erb
```diff
@@ -5,3 +5,130 @@ pip install vspk==5.3.2
...
+boto3==1.6.2
+botocore==1.9.3
+cachetools==3.0.0
+certifi==2018.1.18
+cffi==1.11.5
+chardet==3.0.4
+colorama==0.3.9
+cryptography==2.6.1
+decorator==4.2.1
+deprecation==2.0
```
NickLaMuro: This is the only new package that isn't in the source from `awx`:

https://github.com/ansible/awx/blob/c7bb0f10/requirements/requirements_ansible.txt#L50-L51

kickstarts/partials/post/python_modules.ks.erb
```diff
@@ -5,3 +5,130 @@ pip install vspk==5.3.2
...
+pycparser==2.18
+Pygments==2.2.0
+PyJWT==1.6.0
+pykerberos==1.2.1
+PyNaCl==1.2.1
+pyOpenSSL==17.5.0
+pyparsing==2.2.0
+python-dateutil==2.6.1
+pyvmomi==6.5
+pywinrm==0.3.0
```
NickLaMuro: This one is the only one that has a different notation (`pywinrm[kerberos]==0.3.0`) in the `ansible/awx` file:

https://github.com/ansible/awx/blob/c7bb0f10/requirements/requirements_ansible.txt#L99

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit df94d2e90f0942d9458a831ecd3f4199ca121603
Author: Jason Frey <jfrey@redhat.com>
Date:   Fri Aug 9 14:49:45 2019 -0400

    Merge pull request #341 from carbonin/install_python_modules
    
    Install python modules required for embedded ansible credentials
    
    (cherry picked from commit 441d102a558b8a29df35532cf5ecec21f77083f6)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1734129
```

PR #343, opened 2019-08-20
[Changelog] Update for Sprint 118

JPrause:
[skip ci]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased as of Sprint 118 - 2019-08-19
 
-### Fixed
+### Added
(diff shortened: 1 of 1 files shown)


JPrause:
@miq-bot add_label documentation

CHANGELOG.md
```diff
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 
-## Ivanchuk-Beta2
+## Unreleased as of Sprint 118 - 2019-08-19
+
+### Fixed
```
simaishi: I think this should be `Added`.

JPrause: hmmm,...that PR is labeled as a bug (Fixed). But I can change it to Added.

simaishi: Already discussed with @JPrause but for future reference... Adding those modules was a part of a bug fix. But it just looked strange to put that PR in `Fixed` section when there is no "issue" mentioned and title says "Install modules"...

PR #345, opened 2019-08-29
Add option to use a different npm registry for yarn

simaishi:
Add an option to override npm registry. This will be no-op if npm_registry/npm_registry_cleanup don't exist on the build machine.

Example npm_registry.erb.ks
```
yarn config set registry <your registry>
find ${repos_root} -name yarn.lock -exec sed -i 's|https://registry.yarnpkg.com|<your registry>|g' {} +
find $(gem env gemdir)/bundler -name yarn.lock -exec sed -i 's|https://registry.yarnpkg.com|<your registry>|g' {} +
```

This will cover `yarn install` which is what our UI uses and takes care of all packages except for `yarn`. I will create a follow up PR for `npm install` which will take care of `yarn`.

PR #347, opened 2019-09-03
Fix optional kickstart partial handling and add support for "npm install"

simaishi:
https://github.com/ManageIQ/manageiq-appliance-build/pull/345 worked only partially - when `--local` is used and when `bin/kickstart_build.rb` is used. For nightly build, manageiq-appliance-build is cloned to a different directory and kickstart files from there are used. Since `@build_base` is different in those cases, my previous PR didn't work.

I'm changing to take optional kickstart partials from a different location. Settings like npm registry will be per build machine setting, not per build/branch setting, so it actually makes sense the files live somewhere else that all builds can access.

For the new `render_partial_if_exist` method, use of `Productization.file_for` isn't needed as we can commit all files for downstream builds and this new method won't be used there.

Also added optionally using an override for "npm install" as this was previously possible only for "yarn install".

### DIFF
--- a/kickstarts/partials/post/ui_compile.ks.erb
+++ b/kickstarts/partials/post/ui_compile.ks.erb
@@ -1,8 +1,6 @@
 npm install -g yarn
 
-<% if File.exist?(@build_base.join(KS_PART_DIR, "post", "npm_registry.ks.erb")) %>
-  <%= render_partial "post/npm_registry" %>
-<% end %>
+<%= render_partial_if_exist("/build/optional_kickstart_partials/npm_registry.ks.erb") %>
 
 pushd /var/www/miq/vmdb
   rake update:ui
@@ -20,6 +18,4 @@ popd
 # Clean cache, will be populated again when yarn runs next time
 yarn cache clean
 
-<% if File.exist?(@build_base.join(KS_PART_DIR, "post", "npm_registry_cleanup.ks.erb")) %>
-  <%= render_partial "post/npm_registry_cleanup" %>
-<% end %>
+<%= render_partial_if_exist("/build/optional_kickstart_partials/npm_registry_cleanup.ks.erb") %>
--- a/kickstarts/partials/post/ui_compile.ks.erb
+++ b/kickstarts/partials/post/ui_compile.ks.erb
@@ -1,6 +1,8 @@
+<%= render_partial_if_exist("/build/optional_kickstart_partials/npm_registry.ks.erb") %>
+
 npm install -g yarn
 
-<%= render_partial_if_exist("/build/optional_kickstart_partials/npm_registry.ks.erb") %>
+<%= render_partial_if_exist("/build/optional_kickstart_partials/npm_registry_yarn.ks.erb") %>
 
 pushd /var/www/miq/vmdb
   rake update:ui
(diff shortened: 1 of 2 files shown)


kickstarts/partials/post/ui_compile.ks.erb
```diff
@@ -1,8 +1,6 @@
 npm install -g yarn
 
-<% if File.exist?(@build_base.join(KS_PART_DIR, "post", "npm_registry.ks.erb")) %>
-  <%= render_partial "post/npm_registry" %>
-<% end %>
+<%= render_partial_if_exist("/build/optional_kickstart_partials/npm_registry.ks.erb") %>
```
simaishi: I probably changed the directory name "optional_kickstart_partials" 3-4 times and kept the longest but most descriptive name here... I'm open for suggestions.

simaishi:
@jrafanie please review

kickstarts/partials/post/ui_compile.ks.erb
```diff
@@ -20,6 +20,4 @@ popd
 # Clean cache, will be populated again when yarn runs next time
 yarn cache clean
 
-<% if File.exist?(@build_base.join(KS_PART_DIR, "post", "npm_registry_cleanup.ks.erb")) %>
-  <%= render_partial "post/npm_registry_cleanup" %>
-<% end %>
+<%= render_partial_if_exist("/build/optional_kickstart_partials/npm_registry_cleanup.ks.erb") %>
```
jrafanie: Is it safe to hardcode the `/build` directory?  Is this consistently mapped in all build environments?

jrafanie:
LGTM, other than the minor question

kickstarts/partials/post/ui_compile.ks.erb
simaishi: /build is [our BUILD_BASE](https://github.com/ManageIQ/manageiq-appliance-build/blob/master/scripts/vmbuild.rb#L33), so I think that's the safest place.

jrafanie:
@simaishi can you confirm you did a build with and without `--local`... it looks right but this has no real test coverage

simaishi:
> @simaishi can you confirm you did a build with and without `--local`... it looks right but this has no real test coverage

Tested with and without `--local` as well as calling `bin/kickstart_build.rb`

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 375d1a0284ef57d4928f3ae4ab954f69b1485160
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Tue Sep 3 17:48:57 2019 -0400

    Merge pull request #347 from simaishi/fix_optional_kickstart
    
    Fix optional kickstart partial handling and add support for "npm install"
    
    (cherry picked from commit f5b0c16ffbd14fa8dccbf6c7bedf4f2e1f7f4ffa)
```

simaishi:
Hammer backport details:

```
$ git log -1
commit 33709c02668102419904bca32eff243888d0b563
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Tue Sep 3 17:48:57 2019 -0400

    Merge pull request #347 from simaishi/fix_optional_kickstart
    
    Fix optional kickstart partial handling and add support for "npm install"
    
    (cherry picked from commit f5b0c16ffbd14fa8dccbf6c7bedf4f2e1f7f4ffa)
```

PR #355, opened 2019-09-30
Configure network adapters during kickstart

bdunne:
- Ensure network service is enabled
- Remove invalid configurations for adapters en*
- Add a config for eth0

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1755398

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -91,13 +91,20 @@ $appliance_root/manageiq-setup.sh
 
 <%= render_partial "post/db_init" %>
 
-# make sure we have the device name in all ifcfg-* files
-ls /etc/sysconfig/network-scripts/ifcfg-* | while read FILE
-do
-  # parse the device name from FILE
-  DEVICE=${FILE##*-}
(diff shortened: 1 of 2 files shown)


kickstarts/base.ks.erb
```diff
@@ -91,13 +91,22 @@ $appliance_root/manageiq-setup.sh
...
+DEVICE=eth0
+BOOTPROTO=dhcp
+ONBOOT=yes
+TYPE=Ethernet
+USERCTL=no
+NM_CONTROLLED=no
+DEFROUTE=yes
+EOF
+
+systemctl enable network
```
simaishi: `NetworkManager` is enabled for downstream appliances (see https://github.com/ManageIQ/manageiq-appliance-build/pull/315). So we need both `network` and `NetworkManager` services?

bdunne: Yes, if you set s static IP address and write the config file, it will come up when you call `ifup eth0`, but it won't come up on the next reboot unless the `network` service is enabled.

PR #357, opened 2019-10-03
Copy to devel

bdunne:
:shipit:

### DIFF
--- a/scripts/uploader.rb
+++ b/scripts/uploader.rb
@@ -52,10 +52,29 @@ module Build
         )
 
         puts "Uploading #{appliance} as #{destination_name}...complete: #{destination_url}"
+
+        next unless master?(appliance)
+
+        if nightly?
+          devel = devel_filename(destination_name)
+          puts "Copying   #{appliance} to #{devel}..."
+
+          RestClient::Request.execute(
+            :method  => :copy,
+            :url     => destination_url,
+            :headers => token_headers.merge("Destination" => "/#{container}/#{devel}")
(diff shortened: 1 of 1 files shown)


Fryguy:
@simaishi Please review

Fryguy:
@simaishi I'm going to merge so we can get it into tonight's nightly, but please review anyway.

scripts/uploader.rb
```diff
@@ -114,6 +138,11 @@ def uploaded_filename(appliance_name)
       File.basename(filename)
     end
 
+    def devel_filename(appliance_name)
+      name = appliance_name.split("-")
+      (name[0..1] << "devel").join("-") + File.extname(appliance_name)
```
simaishi: Our GCE image has `.tar.gz` extension `manageiq-gce-master-<date>-<sha>.tar.gz`, so use of `extname` will change to `.gz` and GCE upload won't work.

scripts/uploader.rb
```diff
@@ -52,11 +52,34 @@ def run
...
     end
 
     private
 
+    def master?(filename)
+      filename.include?("-master-")
+    end
+
+    def nightly?
+      !release?
```
simaishi: We do support `type` other than `nightly` and `release` (although I have never personally used it....), so wouldn't it be better to explicitly check for `type == "nightly"`?

This script is currently called only for nightly/release, so the end result will be same, though...

simaishi:
`-stable` next!

PR #370, opened 2019-11-07
Cleanup package installation

simaishi:
- No need to install documentation
- Remove packages installed by default but not needed (`linux-firmware` and `doxygen`)
- Remove `docker` as it's no longer needed
- Remove unused postmodern/ruby-install repo (ruby-install comes from ManageIQ-Master)

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -30,7 +30,7 @@ device mptspi
 
 reboot
 
-%packages
+%packages --excludedocs
 @base --nodefaults
 @development tools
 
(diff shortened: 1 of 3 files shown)


kickstarts/base.ks.erb
```diff
@@ -50,6 +50,9 @@ set -x
 # Show timestamps on log entries for build performance tuning.
 PS4="+ [\t] "
 
+# Don't install doc
+yum-config-manager --setopt=tsflags=nodocs --save
```
simaishi: I don't currently reset this. Do we want to reset or keep `nodocs`?

PR #375, opened 2019-12-05
Build container images in nightly/release build

simaishi:
The new script will build pod images, followed by monolithic image, and push the images.

Since builds for 2 different branches might run at the same time, sources for container images will be cloned to branch specific directories (just like manageiq-appliance-build)

`podman` comes with CentOS 8, but the version of podman on CentOS 8.0 had a few issues, so using `docker-ce` for now. Will revisit when CentOS 8.1 is released (at least 1 issue will be fixed in that version).

cc @carbonin

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,20 @@ Below are instructions on configuring a dedicated build machine to generate appl
     virt-viewer     = Graphical console
     ```
 
+## Setup docker for container build
+
+  * Install docker and start service
+    ```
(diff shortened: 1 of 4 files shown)


simaishi:
@carbonin ready for review

carbonin:
@simaishi does this need any of the npm registry override stuff that you added to pods? Or is that all just set on the build machine with env vars?

simaishi:
All set as the build machine has the env.

PR #381, opened 2020-02-11
CentOS 8 based appliance

simaishi:
Various changes to update our appliances to be CentOS 8 based. A few things to note...
- Not all ovirt-ansible-* packages are available for CentOS 8 and this causes none of the ovirt-ansible-* packages to be installed (due to missing dependency).
- Updated to nodejs 12 as that's the latest LTS release and is available in AppStream cc @himdel
- Due to CentOS 8 not having libssh2-devel (which is needed for rugged gem ssh support), RPMs were built in copr and are installed in %post section. Trying to install in %package fails due to conflicts.

Follow up in separate PRs:
- Change to use ruby from AppStream, rather than using ruby-install. Some extra change will be needed to make this work.
- We should probably update ruby version.

 @jrafanie I'll ping you on those 2 items above.

- Change OS install method/location. We currently use iso stored locally. The build extracts iso contents, modify and re-generate iso to be used during installation. This happens per appliance and it's time consuming. With CentOS 8 iso being 7 GB (CentOS 7 iso is 4.3 GB), this takes about 10 minutes now. I'm trying out different settings to see what works best (speed and reliability).

Other follow up items:
- Patch oz rpm that's in ManageIQ-Build copr repo. Without patch, it still works, but intermediate image files are not sparse copied, causing copy process to take significantly longer (5-7 min vs 20+ min). I've patched oz on our build machine for now.
- Look into oz not handling CentOS 8 config setting. Using CentOS 7 setting works, but it's confusing to see 7 in config/base.tdl when we're actually using 8.

Marking as WIP to do a few more test runs, but ready for review.

### DIFF
--- a/config/base.tdl
+++ b/config/base.tdl
@@ -1,9 +1,9 @@
 <template>
-  <name>CentOS-8.1</name>
+  <name>CentOS-7.4</name>
   <os>
     <rootpw>ozrootpw</rootpw>
-    <name>CentOS-8</name>
-    <version>1</version>
+    <name>CentOS-7</name>
+    <version>4</version>
     <arch>x86_64</arch>
     <install type="iso">
       <iso>file:///build/isos/CentOS-8.1.1911-x86_64-dvd1.iso</iso>
--- a/config/base.tdl
+++ b/config/base.tdl
@@ -1,12 +1,12 @@
 <template>
-  <name>CentOS-8.0</name>
+  <name>CentOS-8.1</name>
   <os>
     <rootpw>ozrootpw</rootpw>
     <name>CentOS-8</name>
-    <version>0</version>
+    <version>1</version>
     <arch>x86_64</arch>
     <install type="iso">
-      <iso>file:///build/isos/CentOS-8-x86_64-1905-dvd1.iso</iso>
+      <iso>file:///build/isos/CentOS-8.1.1911-x86_64-dvd1.iso</iso>
     </install>
   </os>
   <description>CentOS TDL meant to be used with custom kickstart for manageiq.</description>
--- a/config/base.tdl
+++ b/config/base.tdl
@@ -1,15 +1,15 @@
 <template>
-  <name>CentOS-7.4</name>
+  <name>CentOS-8.0</name>
   <os>
     <rootpw>ozrootpw</rootpw>
-    <name>CentOS-7</name>
-    <version>4</version>
+    <name>CentOS-8</name>
+    <version>0</version>
     <arch>x86_64</arch>
     <install type="iso">
-      <iso>file:///build/isos/CentOS-7-x86_64-DVD-1708.iso</iso>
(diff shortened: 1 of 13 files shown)


jrafanie:
> * Change to use ruby from AppStream, rather than using ruby-install. Some extra change will be needed to make this work.
> * We should probably update ruby version.

Depending on how they're packaged, yes, we might have to do extra work.  I'm not sure if that would make development on appliances more difficult if it became more difficult to bundle install in the event we needed packages from app stream to do updates.  Also, we might be limited to the versions they ship which could be ok or not ok depending on how often they do updates.

We can certainly look at this in more depth and see if these are actual problems though.

himdel:
> Updated to nodejs 12 as that's the latest LTS release and is available in AppStream cc @himdel

Thanks, will this be used only for jansa+ or including new builds of ivanchuk?
(I have heard of issues with ivanchuk with node 13, so I may need to investigate if so.)

simaishi:
> Thanks, will this be used only for jansa+ or including new builds of ivanchuk?

Added `ivanchuk/no`. We have no plans to update to node 12 for ivanchuk branch.

simaishi:
@carbonin @Fryguy please review

carbonin:
> Not all ovirt-ansible-* packages are available for CentOS 8 and this causes none of the ovirt-ansible-* packages to be installed (due to missing dependency).

What does this affect specifically? Is this okay for now? @agrare I thought ovirt was using ansible for things like power operations.

agrare:
@carbonin ovirt still uses their ovirtsdk4 gem which is a C extension gem

carbonin:
Okay, so we don't think that anything that we provide "out of the box" will be broken if we're missing the ovirt ansible packages?

simaishi:
> > Not all ovirt-ansible-* packages are available for CentOS 8 and this causes none of the ovirt-ansible-* packages to be installed (due to missing dependency).
> 
> What does this affect specifically? Is this okay for now? @agrare I thought ovirt was using ansible for things like power operations.

Ah, forgot to update that bullet point... I've been trying to get updates/status on those packages for a while, but so far that hasn't been successful. I'm not exactly sure what won't work without it. @dmetzger57 is going to follow up on this. If we want to wait for that, I can put back to WIP.

@dmetzger57 @Fryguy thought?

Fryguy:
I think we can do that in a follow up.  Since this is master only, we still have time, and I'd like to get some usage on CentOS early to work out any other issues.

Fryguy:
This LGTM as is, so I'm going to merge now, and @simaishi can iterate on the rest.

Fryguy:
> Okay, so we don't think that anything that we provide "out of the box" will be broken if we're missing the ovirt ansible packages?

I believe the only thing the ovirt team uses ansible for at the moment is for ovirt cluster upgrade, which is a rarely used feature.

Issue #152, opened 2016-08-23
Cleanup how to add target specific partials for kickstart


simaishi:
Adding partials for specific target can be cleaned up, see https://github.com/ManageIQ/manageiq-appliance-build/pull/150#discussion_r75921955

Also blank lines added for each tag in output can be removed by adding trim mode for better readability.

mfeifer:
@simaishi is this still needed?

mfeifer:
@simaishi does this issue still need to stay open?

simaishi:
We have to keep a list of target that has target specific erb (so it will fail if file is missing), so it's probably best to keep this as is.

PR #383, opened 2020-02-19
Use ERB to generate oz base.tdl

simaishi:
We'll be adding another appliance that will need a different image size, so changed to use ERB to generate base.tdl used by oz.


scripts/vmbuild.rb
```diff
@@ -107,10 +107,12 @@ def verify_run(output)
     imgfac_target = target.imagefactory_type
     ova_format    = target.ova_format
     compression   = target.compression_type
+    image_size    = target.image_size
     $log.info "Building for #{target}:"
 
-    tdl_name = target.name == "azure" ? "base_azure.tdl" : "base.tdl"
-    tdl_file = CFG_DIR.join(tdl_name)
+    tdl_file = CFG_DIR.join("generated", "base_#{target}.tdl")
+    File.write(tdl_file, ERB.new(File.read(CFG_DIR.join("base.tdl.erb")), 0, '-').result(binding))
```
carbonin: What are the extra parameters to `ERB.new` here for? We don't pass those for the kickstart generation.

simaishi: That adds trim mode - this allows removing extra blank lines that get added by using conditions (like <% if xxx %>).  I mentioned in https://github.com/ManageIQ/manageiq-appliance-build/issues/152 for kickstart file, but I never made the change there... 

In case we add conditions later (like using different ISOs based on a flag), I'm just adding with the trim mode now.

Without the trim mode:
```
<% if true %>
line
<% end %>
```
becomes
```

line

```

By using the trim mode with '-':

```
<% if true -%>
line
<% end -%>
```
becomes
```
line
```

PR #385, opened 2020-02-21
Build v2v-conversion-host appliance 

simaishi:
v2v-conversion-host appliance is different from the rest of appliances we build, where kickstart file comes from a different repo and the final appliance file name has different format.

For now, I've opted to handle those differences using `if target == "v2v"`. If we ever need to create another non-ManageIQ appliance, I'll look into either adding new options in Build::Target or re-organizing vmbuild.rb so target specific `if` can be removed.

Requires https://github.com/ManageIQ/manageiq-appliance-build/pull/383

### DIFF
--- a/scripts/cli.rb
+++ b/scripts/cli.rb
@@ -5,14 +5,14 @@ require_relative 'target'
 module Build
   class Cli
     attr_reader :options
-    ALLOWED_TYPES = %w(nightly release test)
-    DEFAULT_TYPE  = "nightly"
-    DEFAULT_REF   = "master"
-    MANAGEIQ_URL  = "https://github.com/ManageIQ/manageiq.git"
-    APPLIANCE_URL = "https://github.com/ManageIQ/manageiq-appliance.git"
-    BUILD_URL     = "https://github.com/ManageIQ/manageiq-appliance-build.git"
-    SUI_URL       = "https://github.com/ManageIQ/manageiq-ui-service.git"
-    V2V_URL       = "https://raw.github.com/ManageIQ/manageiq-v2v-conversion_host-build"
+    ALLOWED_TYPES           = %w(nightly release test)
+    DEFAULT_TYPE            = "nightly"
+    DEFAULT_REF             = "master"
+    MANAGEIQ_URL            = "https://github.com/ManageIQ/manageiq.git"
+    APPLIANCE_URL           = "https://github.com/ManageIQ/manageiq-appliance.git"
+    BUILD_URL               = "https://github.com/ManageIQ/manageiq-appliance-build.git"
+    SUI_URL                 = "https://github.com/ManageIQ/manageiq-ui-service.git"
+    V2V_CONV_HOST_KICKSTART = "https://raw.github.com/ManageIQ/manageiq-v2v-conversion_host-build"
 
     def parse(args = ARGV)
       git_ref_desc   = "provide a git reference such as a branch or tag, non \"#{DEFAULT_REF}\" is required for 'release' type"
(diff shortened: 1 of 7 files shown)


simaishi:
@bdunne @carbonin please review

scripts/cli.rb
```diff
@@ -12,6 +12,7 @@ class Cli
     APPLIANCE_URL = "https://github.com/ManageIQ/manageiq-appliance.git"
     BUILD_URL     = "https://github.com/ManageIQ/manageiq-appliance-build.git"
     SUI_URL       = "https://github.com/ManageIQ/manageiq-ui-service.git"
+    V2V_URL       = "https://raw.github.com/ManageIQ/manageiq-v2v-conversion_host-build"
```
carbonin: This is a bit odd as the others are source an this one is a raw file.

Makes me think, does it make sense for the V2V kickstart file to live in this repo? Why are we building it here while it still exists externally? @Fryguy ?

simaishi: Yeah, I knot it's odd... I thought about cloning first, but I really just need 1 file from there and figured I could just go for a raw file since v2v appliance is different anyway...

Fryguy: Maybe don't name it `_URL`, since it's different?  That is, if it were named V2V_KICKSTART, for example, I'm not sure there would have been much of a question because even by name it's "different"

Fryguy: > Makes me think, does it make sense for the V2V kickstart file to live in this repo? Why are we building it here while it still exists externally? @Fryguy ?

I'm not sure... This isn't part of the main appliance; it's a totally separate appliance.  I'm on the fence where it lives... from a v2v team autonomy point of view, it's easier over there.

Fryguy: @simaishi I know the variable name is getting long, but it's not really a v2v url (when I hear v2v I think of the plugin), but instead this is the conversion host, so maybe V2V_CONV_HOST_KICKSTART ?

scripts/target.rb
```diff
@@ -12,6 +12,7 @@ class Target
       'libvirt'   => ImagefactoryMetadata.new('openstack-kvm', nil, 'qc2', nil, '66'),
       'gce'       => ImagefactoryMetadata.new('gce', nil, 'tar.gz', nil, '66'),
       'ec2'       => ImagefactoryMetadata.new('ec2', nil, 'vhd', 'zip', '66'),
+      'v2v'       => ImagefactoryMetadata.new('openstack-kvm', nil, 'qc2', nil, '10'),
```
Fryguy: Can the target name be v2v_conf_host or something to differentiate that it's not a manageiq appliance?

scripts/uploader.rb
```diff
@@ -22,10 +22,9 @@ def initialize(directory, type)
     def run
       login
 
-      appliances = Dir.glob("#{directory}/manageiq*")
-      appliances.each do |appliance|
-        # Skip badly named appliances, missing git sha: manageiq-openstack-master-201407142000-.qc2
-        unless appliance.match?(/.+-[0-9]{12}-[0-9a-fA-F]+/)
+      Dir.glob("#{directory}/*").each do |appliance|
+        # Skip files without date
+        unless appliance.match?(/.+-[0-9]{12}/)
```
Fryguy: Why are we not checking for the git sha like previously?

scripts/uploader.rb
```diff
@@ -139,21 +138,18 @@ def uploaded_filename(appliance_name)
     end
 
     def devel_filename(appliance_name)
-      name = appliance_name.split("-")
-      extension = ".#{appliance_name.split(".", 2).last}"
-      (name[0..1] << "devel").join("-") + extension
+      name = appliance_name.match(/(.*-).*-[0-9]{8}(?:-\h*)?(.*)/)
+      name[1] + "devel" + name[2]
```
Fryguy: Prefer interpolation over concatenation.  `"#{name[1]}devel#{name[2]}`

scripts/vmbuild.rb
```diff
@@ -155,7 +160,11 @@ def verify_run(output)
     $log.info "Built #{target} with final UUID: #{uuid}"
 
     FileUtils.mkdir_p(destination_directory)
-    file_name = "#{name}-#{target}-#{build_label}-#{timestamp}-#{manageiq_checkout.commit_sha}.#{target.file_extension}"
+    if target.name == "v2v"
+      file_name = "v2v-conversion-host-appliance-#{build_label}-#{timestamp}.#{target.file_extension}"
```
Fryguy: why doesn't this have a commit sha?

scripts/uploader.rb
simaishi: Because v2v appliance won't have SHA in its name, so the only thing guaranteed to be in all images (but not included in non-image files) is the date/time.

If we're to include SHA in the v2v image as well, that will be the SHA of the repo where kickstart file will come from. But the kickstart file will rarely change once we finalize it, so it's kind of misleading to have SHA. The main contents of the image come from v2v-conversion-host container image (it's pulled inside appliance).

Fryguy: The previous check though was to prevent the build name not having a SHA in it

    # Skip badly named appliances, missing git sha: manageiq-openstack-master-201407142000-.qc2

so the new regex will still allow that "badly named appliance".

> But the kickstart file will rarely change once we finalize it, so it's kind of misleading to have SHA.

I guess it depends what the kickstart file is doing... does it pull in some source?  If so, then a SHA (of the code in general, not of just the file) would make sense.

simaishi: > so the new regex will still allow that "badly named appliance".

I've never seen images created without SHA (or any bad names for that matter)... so to me that check was more for filtering out non-image files that exist in that directory, than really filtering out bad named appliances.

I'd hope something will fail earlier if we don't even have the manageiq core repo SHA!

> I guess it depends what the kickstart file is doing... does it pull in some source? If so, then a SHA (of the code in general, not of just the file) would make sense.

The appliance build will pull in container image. The container image (Dockerfile is in the same [repo](https://github.com/ManageIQ/manageiq-v2v-conversion_host-build) as kickstart file - which probably won't change too often, but more often than kickstart file) pulls bunch of RPMs. Probably we can consider v2v-conversion-host-wrapper RPM as the "source"(?) which is in yet another [repo](https://github.com/ManageIQ/manageiq-v2v-conversion_host). But since that's a released RPM, we have to find out the latest version and get the SHA of the repo if we want to go that route.

scripts/cli.rb
simaishi: Changed to V2V_CONV_HOST_KICKSTART

scripts/target.rb
simaishi: Changed to v2v_conv_host

scripts/uploader.rb
simaishi: Done.

PR #386, opened 2020-02-25
Fix date parsing to account for v2v appliance

simaishi:
In https://github.com/ManageIQ/manageiq-appliance-build/pull/385, I missed one place where date parsing needed to be updated. Changed to use the similar regex as generating filenames.


bdunne:
Build is green on Travis, not sure why it isn't updated here.  Merging...

PR #391, opened 2020-03-16
Resize and convert Azure disk

simaishi:
This will eliminate users needing to manually resize/convert Azure disks and closes https://github.com/ManageIQ/manageiq-appliance-build/issues/212.

The standalone fix_azure_disk.rb script was changed to Build::TargetDiskConverter as we don't need to call that script manually now.

The only thing imagefactory `target_image` step does for "hyperv" target type is to run qemu-img to convert to vpc format. Since that conversion is done after resize in TargetDiskConverter, skipping `target_image` step for azure.


scripts/target_disk_converter.rb
```diff
@@ -0,0 +1,46 @@
...
+  class TargetDiskConverter
+    attr_reader :disk
+
+    def initialize(disk)
+      @disk = disk
+    end
+
+    def fix_azure_disk
+      disk_type = AwesomeSpawn.run!("qemu-img info --output json #{disk}").output.match(/format\": "(.*)"/)[1]
+      convert_to_raw if disk_type == "vpc"
```
bdunne: Is it sometimes raw and sometimes vpc?

simaishi: The original script was being used for downstream as well, and upstream was raw and downstream was vpc. We no longer build azure downstream, but I left it, just in case.

PR #392, opened 2020-03-16
Python module update for ansible

simaishi:
Generated the list from the current downstream ansible-tower-venv-ansible RPM using `pip3 freeze -l`

Also removed unnecessary `yum install -y python3-pip`. It's installed as a requirement of python36

Created a new venv and made sure pip install works, not sure if anything else needs to be tested.


carbonin:
I wasn't really intending to track the tower python packages past the initial list.

For example, if we need to upgrade packages that we already ship, that's fine. But I don't see why we would need to add all the same packages that they are.

kickstarts/partials/post/python_modules.ks.erb
```diff
@@ -37,6 +36,7 @@ azure-mgmt-cosmosdb==0.5.2
 azure-mgmt-devtestlabs==3.0.0
 azure-mgmt-dns==2.1.0
 azure-mgmt-hdinsight==0.1.0
+azure-mgmt-iothub==0.7.0
```
carbonin: So something like this, I'm not sure we need. I don't think we'll be managing IOT in Azure.

simaishi:
> I wasn't really intending to track the tower python packages past the initial list.
> 
> For example, if we need to upgrade packages that we already ship, that's fine. But I don't see why we would need to add all the same packages that they are.

I thought we wanted to keep it in sync with downstream, just so it's easier for us to maintain, until someone has time to review what's really needed and remove extra ones. If extra modification should be done from the generated list even now, I'll close this PR so someone else can review and update only what's needed 😄

carbonin:
@Fryguy opinions? @simaishi maybe we can limit the change to just updates so we're not making more work for ourselves when we get around to paring down the list?

Fryguy:
I think I wrote this elsewhere, but I'm seriously considering dropping all of the python dependencies moving forward unless there's something obviously needed, as I'd expect an author to list their dependencies for their playbooks in their galaxy requirements files.

carbonin:
I don't think we can drop *all* of them unless we move to a more generic credential management system.

As long as we're committing to differentiating credential types and providing them in ways specific to a particular module we will need to at least carry those modules and their dependencies.

simaishi:
Based on the comments here, I'm going to close this PR and moving this discussion to https://github.com/ManageIQ/manageiq-appliance-build/issues/423

PR #396, opened 2020-03-18
Bundle with 'systemd' optional group

simaishi:
Related to https://github.com/ManageIQ/manageiq/pull/19874

Also changed to use `bundle config set` rather than `--with` due to deprecation warning:

[DEPRECATED] The `--with` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set with 'qpid_proton'`, and stop using this flag


simaishi:
Jansa backport details:

```
$ git log -1
commit 1b4e91384bf2a1d82666a4b17fd792cf88a1ec4e
Date:   Thu Mar 19 09:10:27 2020 -0400

    Merge pull request #396 from simaishi/bundle_systemd

    Bundle with 'systemd' optional group

    (cherry picked from commit 4e2d7d679bc7fff6078da4a4fe4fdcfe77d411f4)
```

Issue #398, opened 2020-03-19
Missing Doc Step - Place centos8 iso in /build/isos


sethpeterson:
The centos8 iso needs to be placed in /build/isos/ prior to building. This step is missing from the documentation.

Fryguy:
wonder if we can just have a curl command to fetch that?

chessbyte:
@sethpeterson if possible, please make a PR with this fix

simaishi:
README updated in https://github.com/ManageIQ/manageiq-appliance-build/pull/399.

@sethpeterson If you add "Closes \<issue #\>" (like: `Closes https://github.com/ManageIQ/manageiq-appliance-build/pull/399`) in the git commit message in your PR, the referenced issue will be automatically closed by bot when PR is merged. FYI, for your future issues/PRs 😄

PR #399, opened 2020-03-19
Add centos 8 download doc

sethpeterson:
Add documentation step to download and place centos 8 iso.
This includes an example link that is current but that link will change with each release of centos 8. I am not sure if there is a way to link to the latest version directly.

Issue: https://github.com/ManageIQ/manageiq-appliance-build/issues/398

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,13 @@ Below are instructions on configuring a dedicated build machine to generate appl
   initrd http://pxeserver.example.com/sources/centos/8/initrd.img
   boot
   ```
-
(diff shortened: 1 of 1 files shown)


simaishi:
@sethpeterson Thank you for the PR!

The ISO being downloaded should match ISO specified in:
https://github.com/ManageIQ/manageiq-appliance-build/blob/4e2d7d679bc7fff6078da4a4fe4fdcfe77d411f4/config/base.tdl.erb#L9

So I think we're good with this example for now.

PR #371, opened 2019-11-13
Update build machine setup instruction to use CentOS 8

simaishi:
Also cleaned up outdated information.

PR #414, opened 2020-05-04
Remove imagefactory_setup.sh

AlexBaranowski:
Changes:

- add shebang
- set stricter bash behavior
- check and use imagefactory-plugins or imagefactory_plugins - note that this change has backward compatibility
- disable exit on error when making symlinks


simaishi:
@AlexBaranowski We no longer run this script and the step was removed in https://github.com/ManageIQ/manageiq-appliance-build/pull/371. You can just run `imagefactory_dev_setup.sh` in imagefactory/scripts directory now.

I'm sorry I neglected to delete this file in the PR. Would you please convert this PR to remove the file instead?

Issue #148, opened 2016-08-22
Disable ssh password logins


geertj:
Currently both [azure](https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/azure.ks.erb#L8) and [gce](https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/gce.ks.erb#L4) disable password logins in `/etc/ssh/sshd_config`.

Because we have a non-private root password in our appliance, we should really disable network password logons for all platform for obvious security reasons. Note that all our supported platforms have an alternate, secure way of logging into the system:
- Have an agent that allows deploying ssh keys: gce, azure
- Use cloud-init to deploy ssh keys: openstack, ovirt
- Login via standard username/password on not network accessible console: vsphere, ovirt, hyperv, libvirt, vagrant.

Fryguy:
@carbonin Thoughts?

abellotti:
I recall some folks at the design summit asking to be able to ssh to the appliance using their IPA credentials once ext auth is enabled. I wonder if this should be optional.

carbonin:
I think this would be a bit disruptive for installations that are deploying more than a few appliances at a time.

Also I'm not sure if we do this for other things as well, but I know that we rely on scp to retrieve the `v2_key` for environments with multiple appliances. So that, at least, would be broken on first setup.

Maybe something in between would be a better step? Maybe remove ssh for the root user and create a separate user that is usable with ssh but not allowed to `sudo`? Or maybe just make a console option to change the root user password and strongly recommend that is done (MOTD)?

I'm not sure what kind of things people are doing with ssh in the field though. For example if people are using ansible to deploy/configure our appliances (@akrzos), they would likely be quite disrupted by this change. I understand the concern, but we would need to look at this quite a bit more before making this change.

geertj:
Coming back to this after a some more thinking on this. I think the following two statements are correct:

1. Having a user (especially root) with a default password that can login via the network with that password is a real security issue. Many first time users will not know about this, making them vulnerable as they will leave the root password unchanged. And even for more experienced users, there is a window of time, just after deploying a new appliance, and before they change the password, where that appliance is vulnerable. 

[...]

PR #426, opened 2020-06-02
Enforce root password reset at first login

bdunne:
Related to: #148


Fryguy:
@simaishi Please review.

PR #427, opened 2020-06-02
Replace yum with dnf

simaishi:
`dnf` is the default in CentOS 8.


carbonin:
Can we remove hakiri from this repo @Fryguy ? It doesn't seem to work and marks everything red.

Fryguy:
@carbonin I'll take a look

Issue #422, opened 2020-05-29
Move ovirt-ansible-* rpm installation from post script to rpm requires


simaishi:
Since oVirt 4.4 is now GA, installing ovirt-ansible-* rpms can be done through manageiq rpm "requires". It's currently done in post script: https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/ovirt_ansible.ks.erb

3 things need to be done:

- Update rpm 'requires': https://github.com/ManageIQ/manageiq-rpm_build/pull/58
- Remove from appliance kickstart: https://github.com/ManageIQ/manageiq-appliance-build/pull/428
- Remove from manageiq-base image (https://github.com/ManageIQ/manageiq-pods/pull/548)

simaishi:
@Fryguy this is ready for review

PR #428, opened 2020-06-02
Move ovirt-ansible-roles to manageiq-gemset requires

simaishi:
A part of https://github.com/ManageIQ/manageiq-appliance-build/issues/422

Requires https://github.com/ManageIQ/manageiq-rpm_build/pull/58 (and new rpm build with this change)

### DIFF
--- a/kickstarts/partials/post/repos.ks.erb
+++ b/kickstarts/partials/post/repos.ks.erb
@@ -8,6 +8,8 @@ dnf config-manager --setopt=epel.exclude=*qpid-proton* --save
 dnf config-manager --enable manageiq-*-nightly
 <% end %>
 
+dnf install -y https://resources.ovirt.org/pub/yum-repo/ovirt-release44.rpm
+
 pushd /etc/yum.repos.d/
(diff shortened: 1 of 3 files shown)


kickstarts/partials/post/repos.ks.erb
```diff
@@ -8,6 +8,8 @@ dnf config-manager --setopt=epel.exclude=*qpid-proton* --save
 dnf config-manager --enable manageiq-*-nightly
 <% end %>
 
+dnf install -y https://resources.ovirt.org/pub/yum-repo/ovirt-release44.rpm
```
bdunne, line 11: Does this need to happen before the `%post` section in order for https://github.com/ManageIQ/manageiq-rpm_build/pull/58/files#diff-a50ce4f4455f4c6ef278a7ec5b3ab2f6R29 to work?

simaishi: No, ovirt-ansible-roles and its dependencies are all coming from existing repos specified in kickstarts/partials/main/repos.ks.erb.

Issue #429, opened 2020-06-03
Include a dependency manifest in the build


Fryguy:
@simaishi As discussed, we already have the Ruby gems list, but it would be nice to have python, node, and even system packages in the manifest.

- [x] ruby gems (https://github.com/ManageIQ/manageiq-rpm_build/pull/64)
- [x] npm packages (https://github.com/ManageIQ/manageiq-rpm_build/pull/64)
- [x] webpack (https://github.com/ManageIQ/manageiq-rpm_build/pull/75)
- [x] python modules (https://github.com/ManageIQ/manageiq-rpm_build/pull/108)
- [x] rpm (https://github.com/ManageIQ/manageiq-appliance/pull/289, https://github.com/ManageIQ/manageiq-rpm_build/pull/73, #431 and https://github.com/ManageIQ/manageiq-pods/pull/563)

Fryguy:
I'd like the manifest to have (if possible): package, version, license, source

where `source` is the source of the package.  In the case of system packages, source would be which yum repository it came from.  For something like Ruby, it might be things like rubygems.org vs rubygems.manageiq.org vs a git URL.  As this might be difficult to get, I'm ok with this being missing.

Fryguy:
Sounds good to me @simaishi

bdunne:
Anything else required before closing this issue?

PR #431, opened 2020-06-16
Create rpm manifest

simaishi:
Requires https://github.com/ManageIQ/manageiq-appliance/pull/289 and https://github.com/ManageIQ/manageiq-rpm_build/pull/73

A part of #429

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -132,6 +132,8 @@ echo "" >> /etc/ssh/sshd_config
 # Remove packages not needed
 dnf -C -y --noplugins remove linux-firmware doxygen
 
+/usr/bin/generate_rpm_manifest.sh
+
 # CentOS 7.3 isn't clearing out spaces during install.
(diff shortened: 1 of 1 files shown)


Fryguy:
Would it make more sense to provider this as a script with the source and then call it from kickstart and podified (so we don't have to duplicate the code?)

simaishi:
> Would it make more sense to provide this as a script with the source and then call it from kickstart and podified (so we don't have to duplicate the code?)

You're right. Moved to https://github.com/ManageIQ/manageiq-appliance/pull/289

simaishi:
Jansa backport details:

```
$ git log -1
commit 72ef3b96401e7a8f1a8213d0fcc9554b43605908
Date:   Wed Jul 1 13:51:33 2020 -0400

    Merge pull request #431 from simaishi/rpm_manifest

    Create rpm manifest

    (cherry picked from commit c1a86324cccf128ed0fbc3943977415996c11ec3)
```

PR #440, opened 2020-09-14
Allow ansible venv to access system packages

simaishi:
Currently, setting up ansible venv is failing:

```
    Running setup.py install for pycurl: started
    Running setup.py install for pycurl: finished with status 'error'
    ....
    src/pycurl.h:174:13: fatal error: openssl/ssl.h: No such file or directory
     #   include <openssl/ssl.h>
                 ^~~~~~~~~~~~~~~
```
Looks like a newer version of pycurl (7.43.0.6) requires openssl-devel now. But what's required in venv is `Collecting pycurl>=7.19.0` and we already have `python3-pycurl-7.43.0.2-4.el8.x86_64` installed.  There are more packages like that where we already have the needed version installed by rpm but we're trying to pip install in venv again.

So I'm adding the flag to allow venv to be able to access system packages to eliminate duplicates and avoid running into errors like this.

### DIFF
--- a/kickstarts/partials/post/python_modules.ks.erb
+++ b/kickstarts/partials/post/python_modules.ks.erb
@@ -15,7 +15,7 @@ pip3 install virtualenv
 
 mkdir -p /var/lib/manageiq/
 pushd /var/lib/manageiq
-  virtualenv venv
+  virtualenv --system-site-packages venv
   source venv/bin/activate
 
   cat > requirements.txt << EOF


simaishi:
cc @NickLaMuro

NickLaMuro (approved):
I think this is probably fine, but thanks for pinging me on this.  I will keep it in mind going forward if there are issues with the python packages that come from the python RPM packages s giving strange errors when run with `EmbeddedAnsible`, as it could be related to this change (hopefully that won't happen though).

simaishi:
Jansa backport details:

```
$ git log -1
commit 9d4b959db6424e0690055c0b83ecea5fef27e094
Date:   Tue Sep 15 10:32:18 2020 -0400

    Merge pull request #440 from simaishi/python_system_package

    Allow ansible venv to access system packages

    (cherry picked from commit 7f4b839a006dba37d6297c918bdd44d43825ccf1)
```

PR #458, opened 2020-12-09
Add an option to delete artifacts after upload

simaishi:
This is useful when running builds on an off-premise machine as artifacts are no longer needed on the machine once uploaded.

Need #457 to actually make a use of this in nightly builds.

### DIFF
--- a/scripts/spec/uploader_spec.rb
+++ b/scripts/spec/uploader_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 require 'uploader'
 
 describe Build::Uploader do
-  subject { described_class.new("directory", "type") }
+  subject { described_class.new("directory", "type", false) }
 
   it "#devel_filename" do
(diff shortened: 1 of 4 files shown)


scripts/spec/uploader_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'uploader'
 
 describe Build::Uploader do
-  subject { described_class.new("directory", "type") }
+  subject { described_class.new("directory", "type", "delete") }
```
bdunne, line 5: Shouldn't `"delete"` be a boolean instead?

simaishi: You're right. That's not used in the spec, but should match. Fixed.

bdunne (approved):

scripts/spec/uploader_spec.rb
bdunne: Thanks!

PR #462, opened 2021-02-23
Reduce libvirt differences

bdunne:
- Allow Openstack IPv6
- Allow GlusterFS SSA from any appliance image


kickstarts/partials/packages/includes.ks.erb
```diff
@@ -6,6 +6,7 @@ manageiq-release
 
 createrepo
 fuse-libs
+glusterfs-fuse
```
simaishi, line 9: Since this is no longer appliance specific, this should probably move to manageiq-core or manageiq-gemset rpm spec file.

kickstarts/partials/packages/includes.ks.erb
```diff
@@ -20,11 +21,8 @@ WALinuxAgent
 dracut-config-generic
 <% when "hyperv" %>
 hyperv-daemons
-<% when "libvirt" %>
-qemu-guest-agent
-<% when "ovirt" %>
+<% when "libvirt", "ovirt" %>
```
Fryguy, line 24: ovirt goes away now, right?

kickstarts/partials/packages/includes.ks.erb
Fryguy: Should we add a comment that this is for "SSA on ovirt" ?

bdunne: https://github.com/ManageIQ/manageiq-rpm_build/pull/139

kickstarts/partials/packages/includes.ks.erb
bdunne: Not yet, there's still a few things that need to be done before we can remove the build.  I'm trying to test a libvirt image in ovirt, but having trouble getting the image uploaded

Fryguy:
Merged https://github.com/ManageIQ/manageiq-rpm_build/pull/139

PR #479, opened 2021-05-26
[WIP] introduce admin user

kbrock:
The new manageiq user needs access to the v2_key -- so when the key is created, the
group is properly set.

This code is not fully tested. I need a little more work until I am able to build appliances on my local box and test the actual implementation of the kickstart, but this is what I typed on the command line.

To be honest, I'm not sure if the proper setup here is to chgrp the v2_key or to `chmod 2775` the directory to help others in the future with their `v2_key` endeavors.

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -17,6 +17,8 @@ keyboard us
 
 # Root password: smartvm
 rootpw  --iscrypted $1$DZprqvCu$mhqFBjfLTH/PVvZIompVP/
+user --name=manageiq --groups=wheel --iscrypted --password $1$DZprqvCu$mhqFBjfLTH/PVvZIompVP/
+usermod -a -G manageiq root
 
 authconfig --enableshadow --passalgo=sha512
 selinux --enforcing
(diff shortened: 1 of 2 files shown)


agrare:
@kbrock will we also need to do this (add user and chgrp) in the rpm build for existing appliances?

kbrock:
sounds like we should move all of this over to the rpms

kbrock:
the rpm now has the manageiq user.
Wondering if we want to transition the appliance to using the manageiq user or another one. I'd actually prefer `admin` rather than using `manageiq` but that may be introducing too many special accounts.

Thinking we don't want our service user to be part of the `wheel` group. But the user that logs into this box needs to be.

chessbyte:
> Wondering if we want to transition the appliance to using the manageiq user or another one. I'd actually prefer admin rather than using manageiq but that may be introducing too many special accounts.

Maybe `manageiq-admin`?

kbrock:
I like `admin`

The user will `sudo` as root (or user `manageiq`) to do most of their configuration

Thinking of just closing this for now.

PR #487, opened 2021-09-16
Don't systemctl enable evm-watchdog

agrare:
evm-watchdog was removed and replaced with a standard Restart=on-failure for evmserverd and manageiq-db-ready

https://github.com/ManageIQ/manageiq-appliance/pull/327


jrafanie:
✂️ 🔥 ✂️ 🔥 ✂️ 🔥 

🍰  🍪 👏 🙇 😍 🎉

Issue #158, opened 2016-09-14
Code org: move build code into build directory, rename scripts to lib


jrafanie:
Seen in #154 

Note, we really need to move files to the right places and re-namespace things based on standard ruby conventions:

```
vmbuild.rb             -> build.rb
cli.rb                 -> build/cli.rb
git_checkout.rb        -> build/git_checkout.rb
kickstart_generator.rb -> build/kickstart_generator.rb
productization.rb      -> build/productization.rb
target.rb              -> build/target.rb
productization.rb      -> build/productization.rb
uploader.rb            -> build/uploader.rb
```

Also, rename scripts directory to lib.

PR #507, opened 2022-02-10
Switch to GitHub Actions

Fryguy:
Switch to GitHub Actions

[CI](https://github.com/Fryguy/manageiq-appliance-build/actions/workflows/ci.yaml)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby-version:
+        - '2.6'
+        - '2.7'
+    env:
+      BUNDLE_PATH: scripts/vendor/bundle
```
Fryguy, line 14: This whole scripts thing is just weird.  See also https://github.com/ManageIQ/manageiq-appliance-build/issues/158

chessbyte: assuming that is for future PRs?

Fryguy: yeah, this is too complicate to solve here.

PR #511, opened 2022-05-23
Don't automatically initialize appliances

agrare:
For demo purposes it is nice to have an appliance which automatically initializes itself on boot, but this makes it quite difficult to use these images in a production deployment which requires multiple appliances.

This removes the auto-initialization on first-boot allowing the user to use appliance_console to connect the appliance to an existing region or create a new one.

Follow-up:
* Remove `manageiq-initialize.service` from manageiq-appliance - https://github.com/ManageIQ/manageiq-appliance/pull/359
* Document configuring kafka - https://github.com/ManageIQ/manageiq-documentation/pull/1675
* Reduce the root disk size for the image builds

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -23,7 +23,6 @@ timezone --utc America/New_York
 
 <%= render_partial "main/bootloader" %>
 <%= render_partial "main/disk_layout" %>
-<%= render_partial "main/db_fs" %>
 
 module --name mod_auth_openidc
 module --name ruby --stream 2.7
--- a/kickstarts/partials/main/db_fs.ks.erb
+++ b/kickstarts/partials/main/db_fs.ks.erb
@@ -1,3 +1,0 @@
-part pv.2             --ondrive=vda                     --size=10240 --grow
-volgroup vg_data      --pesize=4096 pv.2
-logvol /var/lib/pgsql --name=lv_pg --vgname=vg_data --size=8192 --grow --fstype=xfs
--- a/kickstarts/partials/post/appliance_init.ks.erb
+++ b/kickstarts/partials/post/appliance_init.ks.erb
@@ -1,1 +1,0 @@
-systemctl enable manageiq-initialize
--- a/scripts/spec/target_spec.rb
+++ b/scripts/spec/target_spec.rb
@@ -25,7 +25,7 @@ describe Build::Target do
   end
 
   it "#image_size" do
-    expect(described_class.new("openstack").image_size).to eql "66"
+    expect(described_class.new("openstack").image_size).to eql "58"
   end
 
   it "#sort" do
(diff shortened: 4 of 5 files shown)


Fryguy:
We use to use manageiq-initialize to initialize kafka as well.  What would we do in that case in the future?

Fryguy:
See also https://github.com/ManageIQ/manageiq-documentation/pull/1626

Fryguy:
Ok that's what I thought...we will just need to document these steps properly.  Thinking ahead, this may also help us with database upgrades, which we've been having trouble with.

kbrock:
Encouraging people to setup an external database would be huge for us.
We do this anyways for most larger deploys, and in the container world, it is becoming more and more standard.

That and gem/python dependencies are among our last major hurdles for upgrades.

agrare:
@bdunne yup I have that in the TODO list above, `Reduce the root disk size for the image builds`

agrare:
And immediately after the database is create, the `evmserverd.service` is activated and `manageiq-db-ready.service` runs

bdunne:
I don't know where 44 came from.  It looks like it's been 66 (61 for azure) for a long time upstream.

agrare:
So I did a little bit of digging on the disk_size issue.

The disk size was originally 40GB: https://github.com/ManageIQ/manageiq-appliance-build/commit/b605f054d71da90d3b822a5bbecd66f0fd957a98#diff-0d2bf1360c948bfb13e78918345353260ed0d447224e0210328f76b9b605f639R14

Then it was increased to 50G to account for the log volume: 3fb352885f3b5c3d96adbc4173d237adeab0353c

And then 66G for embedded ansible when it was using AWS: 911759d3779560814dc5f558bbdcd252ed56a045

Other downstream products also had the root disk size of 66G and those expected the database to be on a secondary disk.

I think we can probably drop back down to 50G (and update the docs accordingly) but the size doesn't appear to be correlated to having a database on the main disk so I think we can leave this for a follow-up, thoughts?

agrare:
@bdunne this is why azure is smaller: https://github.com/ManageIQ/manageiq-appliance-build/pull/128

bdunne:
The 8GB for the database was included in the original 40GB https://github.com/ManageIQ/manageiq-appliance-build/commit/b605f054d71da90d3b822a5bbecd66f0fd957a98#diff-fb329bbe17b199ccc8278c9f0f61b8aeb558a411c1095f7c989fb9c519468827R36-R37 so I think we should drop the disk size by that much if we're going to remove the lv.

agrare:
@bdunne so do you want to drop it by 8GB here for the database then separately drop it by 16GB for the removal of AWX?

bdunne:
> @bdunne so do you want to drop it by 8GB here for the database then separately drop it by 16GB for the removal of AWX?

👍🏻 Sounds good

agrare:
@bdunne done

PR #524, opened 2022-09-20
Update appliances to use ruby 3

jrafanie:
~~Depends on ruby 3 support from https://github.com/ManageIQ/manageiq/pull/21531~~ MERGED 🎉 
Depends on https://github.com/ManageIQ/manageiq-rpm_build/pull/304 for ruby 3 rpm builds


jrafanie:
I assume this would work since we tested images for podified which included the necessary rpm_build image changes.  See https://github.com/ManageIQ/manageiq-rpm_build/pull/304

PR #527, opened 2022-09-28
Avoid installing weak dependencies

Fryguy:
Similar change to https://github.com/ManageIQ/manageiq-pods/pull/846

@agrare Please review.  I'm not sure this will actually fix the python 3.9 issue, but we can see after build.


agrare (approved):

agrare:
@Fryguy should this go back to `oparin` ?

Fryguy:
I want to see what a master build looks like first, but I'm guessing yes.

PR #528, opened 2022-09-28
Ensure psutil is installed for all pythons

Fryguy:
@agrare Please review.

### DIFF
--- a/kickstarts/partials/post/python_modules.ks.erb
+++ b/kickstarts/partials/post/python_modules.ks.erb
@@ -1,7 +1,10 @@
-dnf install -y python38-pip python38-devel libcurl-devel libffi-devel libxml2-devel libxslt-devel krb5-devel openssl-devel gcc make
+# Find all intalled pythons and be sure psutil is installed in all of them
+psutils=$((for p in $(find /usr/lib -type d -name python3.* | grep -v python3.6); do basename $p | tr -d '.' | echo "$(cat -)-psutil"; done) | xargs echo)
+
+dnf install -y $psutils python38-pip python38-devel libcurl-devel libffi-devel libxml2-devel libxslt-devel krb5-devel openssl-devel gcc make
 
 # For embedded ansible
-pip3.8 install virtualenv psutil==5.6.6
+pip3.8 install virtualenv
 
 mkdir -p /var/lib/manageiq/
 pushd /var/lib/manageiq


Fryguy:
Not my finest work, but it works 🤷 .

We're gonna have to do this very different with a proper rpm if we want upgrades to work properly, but this will get us over the hump.

agrare:
@Fryguy so removing the weak dependencies didn't remove python39?

Fryguy:
No.  Turns out 3.9 is a dep on that version of ansible.  Still not sure why appliances are a different version of ansible than podified

agrare:
Interesting, and ansible/ansible-core are slightly different versions on podified and appliances right?

kickstarts/partials/post/python_modules.ks.erb
```diff
@@ -1,7 +1,10 @@
-dnf install -y python38-pip python38-devel libcurl-devel libffi-devel libxml2-devel libxslt-devel krb5-devel openssl-devel gcc make
+# Find all intalled pythons and be sure psutil is installed in all of them
+psutils=$((for p in $(find /usr/lib -type d -name python3.* | grep -v python3.6); do basename $p | tr -d '.' | echo "$(cat -)-psutil"; done) | xargs echo)
+
+dnf install -y $psutils python38-pip python38-devel libcurl-devel libffi-devel libxml2-devel libxslt-devel krb5-devel openssl-devel gcc make
```
agrare, line 4: If we are going to loop through all pythons I wonder if we need to install `python*-devel` as well?  Would `dnf install python39-psutil` fail if `python39-devel` weren't installed?

Fryguy: No, python39-psutil is standalone and doesn't require python39-devel.

Fryguy: python38-devel is only needed to build some of the eggs we install in our venv.

agrare (approved):

Fryguy:
> Interesting, and ansible/ansible-core are slightly different versions on podified and appliances right?

yeah
- appliance: 
   ```
   ansible-6.0.0-1.el8.next.noarch
   ansible-core-2.13.3-1.el8.x86_64
   ```

- podified:
   ```
   ansible-5.4.0-3.el8.noarch
   ansible-core-2.12.7-1.el8.x86_64
   ```

Fryguy:
on appliance, ansible is coming from `@epel-next`, ansible-core is coming from `@appstream`

```
epel-next,Extra Packages for Enterprise Linux 8 - Next - x86_64,/etc/yum.repos.d/epel-next.repo,rsync://ftp-osl.osuosl.org/fedora-epel/next/8/Everything/x86_64/ (143 more),,https://mirrors.fedoraproject.org/metalink?repo=epel-next-8&arch=x86_64&infra=stock&content=centos
```

but on podified, ansible and ansible-core are coming from `@System` and `@epel-next` doesn't exist.

PR #443, opened 2020-10-13
[WIP] Remove post script to set up ansible venv

simaishi:
ansible venv is now a standalone rpm.

Depends on:
*  Build manageiq-ansible-venv rpm for supported architectures
* Add 'Requires' to rpm spec file (https://github.com/ManageIQ/manageiq-rpm_build/pull/111)
* Remove from manageiq-pods build (https://github.com/ManageIQ/manageiq-pods/pull/636)

For #423

PR #534, opened 2022-10-25
Remove post script to set up ansible venv, moved to rpm build

Fryguy:
Rebased / Updated version of #443 

Depends on
- [x] https://github.com/ManageIQ/manageiq-pods/pull/906


Fryguy:
@bdunne This is ready to go

PR #409, opened 2020-04-07
Control qpid-proton rpm version by installing from ManageIQ copr repo

simaishi:
Disallow installing qpid-proton from epel repo (so it will come from ManageIQ copr repo), to ensure the version matches with the gem version.

Related to: https://github.com/ManageIQ/manageiq/issues/19942 and https://github.com/ManageIQ/manageiq/pull/20016

Building 0.30.0 rpm in ManageIQ-Master and ManageIQ-Jansa copr repo now.  Until https://github.com/ManageIQ/manageiq/pull/20016 is merged, bundle will fail on master/jansa appliances with or without this PR.

PR #559, opened 2023-12-01
Use el8/9 epel qpid proton 0.37.0+

jrafanie:
Undo https://github.com/ManageIQ/manageiq-appliance-build/pull/409 now that we're ready to use the upstream packages in el8 or el9 epel qpid proton 0.37.0+

To be merged with https://github.com/ManageIQ/manageiq/pull/22271
Needed for https://github.com/ManageIQ/manageiq/issues/22696

Followup: https://github.com/ManageIQ/manageiq-appliance-build/issues/560

### DIFF
--- a/kickstarts/partials/main/repos.ks.erb
+++ b/kickstarts/partials/main/repos.ks.erb
@@ -2,7 +2,7 @@
 repo --name=baseos     --baseurl=http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/
 repo --name=appstream  --baseurl=http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/
 repo --name=extras     --baseurl=http://mirror.centos.org/centos/8-stream/extras/x86_64/os/
-repo --name=epel       --mirrorlist=https://mirrors.fedoraproject.org/mirrorlist?repo=epel-8&arch=x86_64 --excludepkgs=*qpid-proton*
+repo --name=epel       --mirrorlist=https://mirrors.fedoraproject.org/mirrorlist?repo=epel-8&arch=x86_64
 
 repo --name=manageiq-18-radjabov         --baseurl=https://rpm.manageiq.org/release/18-radjabov/el$releasever/$basearch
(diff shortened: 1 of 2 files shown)


Fryguy:
Is it possible to put the `dnf config-manager --setopt=epel.exclude= --save` in a package that would get installed first, or is there no guarantee to the order?

Fryguy:
> > Is it possible to put the dnf config-manager --setopt=epel.exclude= --save in a package that would get installed first, or is there no guarantee to the order?
> Package dependencies will guarantee the order but I think that DNF resolves the requirements before actually installing any package, but I can test it

This just got me thinking that it might be interesting to have an "installer" RPM.  So you'd dnf upgrade the installer, then run the installer, which could run any "pre-install" steps, then it would run dnf upgrade on the rest.  We were talking about having a script that the user could run beforehand, but this would simplify even that script step.

cc @bdunne

jrafanie:
> This just got me thinking that it might be interesting to have an "installer" RPM. So you'd dnf upgrade the installer, then run the installer, which could run any "pre-install" steps, then it would run dnf upgrade on the rest. We were talking about having a script that the user could run beforehand, but this would simplify even that script step.
> 
> cc @bdunne

wouldn't you still have a dependency tree of

```
installer rpm
|
 \
  - manageiq-*
  |
  - manageiq-gemset
     \
      - qpid-proton-c >= 0.37.0
```

If manageiq* doesn't depend on the installer, you can upgrade the manageiq* rpms without the installer.  I'm not sure how you'd force it to be in sync.

If you have the tree above, you can't resolve the qpid-proton-c dependency until you run the command to clear the dnf epel exclusion.  I think it's a chicken and egg problem.

Fryguy:
@jrafanie As discussed, no, the idea is the installer/upgrader rpm would be independent.  Then any "special" pre-install stuff could be put in a script.  In fact, since it's pre-install, we would also know what version we are coming from and can make different choices.

I will open a separate issue to discuss the details - this will be for radjabov anyway.

PR #563, opened 2024-02-21
Upgrade to el9

bdunne:

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -3,7 +3,7 @@
 # By default run the text based installer to see dependency info in screenshots
 <%= ENV['KS_GUI'] ? "" : "text" %>
 
-logging --level=debug
+logging
 
 #version=DEVEL
 cdrom
@@ -15,7 +15,7 @@ keyboard us
 <%= render_partial "main/network" %>
 
 # Root password: smartvm
-rootpw  --iscrypted $1$DZprqvCu$mhqFBjfLTH/PVvZIompVP/
+rootpw --allow-ssh --iscrypted $1$DZprqvCu$mhqFBjfLTH/PVvZIompVP/
 
 selinux --enforcing
 timezone --utc America/New_York
@@ -23,10 +23,7 @@ timezone --utc America/New_York
 <%= render_partial "main/bootloader" %>
 <%= render_partial "main/disk_layout" %>
 
-module --name mod_auth_openidc
-module --name ruby --stream 3.0
-module --name postgresql --stream 13
-
+module --name ruby --stream 3.1
 reboot
 
 %packages --excludedocs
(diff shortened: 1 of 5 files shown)


kickstarts/base.ks.erb
```diff
@@ -111,7 +108,7 @@ echo "" >> /etc/ssh/sshd_config
 dnf -C -y --noplugins remove linux-firmware doxygen
 
 # Ensure packages are up to date
-dnf -y update --nobest
```
bdunne, line 114: We shouldn't need the `--nobest`.  I think it was added for an ansible rpm issue, but we're on a much newer version now and there shouldn't be any dependency issues anymore.  (I didn't see any issues during install or update of my test appliance)

scripts/vmbuild.rb
```diff
@@ -39,7 +39,7 @@
 STORAGE_DIR         = BUILD_BASE.join("storage")
 ISOS_DIR            = Pathname.new("/build/isos") # NOTE: We are not using run-specific BUILD_BASE in order to share the source iso across builds
 IMAGES_DIR          = BUILD_BASE.join("images")
-ISO_FILE            = ISOS_DIR.glob("CentOS-Stream-8-x86_64-*-dvd1.iso").sort.last.expand_path
+ISO_FILE            = ISOS_DIR.glob("CentOS-Stream-9-*-x86_64-dvd1.iso").sort.last.expand_path
```
bdunne, line 42: I tried to quickly to change this to the much smaller boot ISO, but that didn't work.  I'll open an issue to investigate that.  It could shave 10-20 minutes per image off of the build time.

kickstarts/centos8_build_machine.ks
```diff
@@ -1,4 +1,4 @@
-#version=RHEL8
+#version=RHEL9
```
bdunne, line 1: We don't use this file right now, but I'd like to keep it as a reference for setting up new build machines.

kickstarts/centos8_build_machine.ks
```diff
@@ -7,10 +7,10 @@ zerombr
 clearpart --all --initlabel --drives=sda
 partition /boot --ondisk=sda --asprimary --size=1024 --fstype=xfs
 partition pv.1 --ondisk=sda --asprimary --size=10240 --grow
-volgroup kegerator8 pv.1
-logvol swap --vgname=kegerator8 --name=swap --size=8192
-logvol / --vgname=kegerator8 --name=root --size=20480 --fstype=xfs
-logvol /build --vgname=kegerator8 --name=build --size=20480 --fstype=xfs --grow
+volgroup kegerator pv.1
```
Fryguy, line 10: This name is so silly haha - can we pick any volgroup name or does it have to be this?

kickstarts/centos8_build_machine.ks
```diff
@@ -38,7 +38,7 @@ logging --level=debug
 # Network information
 network --bootproto=dhcp --device=eth1 --onboot=off --noipv6 --no-activate # CHANGEME or remove based on your hardware
 network --bootproto=static --device=eth0 --gateway=192.0.2.1 --ip=192.0.2.2 --nameserver=192.0.2.1 --netmask=255.255.252.0 --noipv6 --activate # CHANGEME
-network --hostname=kegerator8.example.com # CHANGEME
+network --hostname=kegerator.example.com # CHANGEME
```
Fryguy, line 41: same here?

scripts/vmbuild.rb
bdunne: https://github.com/ManageIQ/manageiq-appliance-build/issues/564

kickstarts/centos8_build_machine.ks
bdunne: I decided to remove changes to this file from this PR

Fryguy (approved):

PR #567, opened 2024-04-03
[quinteros] Revert "Changes after new branch quinteros"

bdunne:
This reverts commit 5a12806e7f9f063ea4c0868ec186ae64a94954ca.


Fryguy:
Is there a way we could find these problems with specs. For example we know the q branch is 18 because q is the 18th letter so maybe a spec around the name to the value in these fields?

PR #568, opened 2024-04-10
Add appliance smoke tests

bdunne:
Running this will:
- Deploy a VM
- Change the default password (since we enforce that)
- Set the hostname
- Use appliance_console to configure
  - encryption key
  - database
  - messaging
  - start evmserverd
- run all tests in the smoke_test directory
- shut down the VM and clean up all files created.

### DIFF
--- a/scripts/smoke_test.sh
+++ b/scripts/smoke_test.sh
@@ -1,0 +1,292 @@
+#!/bin/bash
+
+if [[ -z $1 ]]; then
+  echo "Missing test image!"
+  exit 1
+fi
+
+image_name=$(basename $1)
+echo "=== Smoke testing ${image_name}"
+
+[[ "$image_name" =~ ^(.+)-(libvirt|ovirt)-([0-9]*)([a-z]*)(.+)\.qc2$ ]]
+product_name=${BASH_REMATCH[1]}              #manageiq
+platform=${BASH_REMATCH[2]}                  #libvirt
+product_version_number=${BASH_REMATCH[3]}    #18? (this is almost always blank, will fill in later)
+product_version_name=${BASH_REMATCH[4]}      #quinteros?
+product_version_remainder=${BASH_REMATCH[5]} #-1-20240403
+# Fill in product version number if missing
+if [[ -z "$product_version_number" ]]; then
+  letters=({a..z})
+  letter=${product_version_name::1}
+  for (( i=0; ; i++ ))
+  do
+    if [[ ${letters[$i]} == $letter ]]; then
+      product_version_number=$(($i + 1))
+      break
+    fi
+  done
+fi
+
+script_directory=$(dirname -- "$( readlink -f -- "$0"; )")
+
+vm_name="smoketest-${product_version_number}${product_version_name}${product_version_remainder}"
+newpassword=smartvm123
+
+echo "=== Copying image..."
+disk_image="/var/lib/libvirt/images/${vm_name}.qc2"
+cp $1 $disk_image
+
+echo "=== Creating database disk"
+db_disk="/var/lib/libvirt/images/${vm_name}-db.qc2"
+qemu-img create -f qcow2 $db_disk 2G
+
+echo "=== Creating VM..."
+virt-install -n $vm_name --memory 4096 --vcpus 2 --cpu host --disk $disk_image --disk $db_disk --network default --graphics spice --osinfo centos-stream9 --import --noautoconsole
+
+echo "=== Waiting for VM to boot..."
+sleep 30
+
+echo "=== Finding IP Address..."
+ip_address=$(virsh guestinfo $vm_name --interface | grep 192.168.122. | cut -f2 -d: | xargs)
+echo "--- Found ${ip_address}"
+
+echo "=== Waiting for SSH..."
+while true; do
+  ncat ${ip_address} 22 < /dev/null && break
+  sleep 1
+done
+sleep 1
+
+echo "=== Adding to known_hosts..."
+cat ~/.ssh/known_hosts | grep -v $ip_address > ~/.ssh/known_hosts
+ssh-keyscan $ip_address >> ~/.ssh/known_hosts
+
+echo "=== Changing default password..."
+/usr/bin/expect <<EOF;
+spawn sshpass -p smartvm ssh -tt root@$ip_address true
+match_max 100000
+expect "New password: "
+send -- "$newpassword\r"
+expect "Retype new password: "
+send -- "$newpassword\r"
+expect eof
+EOF
+
+
+### Configure the appliance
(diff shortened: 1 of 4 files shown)


scripts/smoke_test.sh
```diff
@@ -0,0 +1,258 @@
...
+
+if [[ $1 == "" ]]; then
+  echo "Missing test image!"
+  exit 1
+fi
+
+image_name=$(basename $1)
+echo "=== Smoke testing ${image_name}"
+
+[[ $image_name =~ ^(.+)-(libvirt|ovirt)-([0-9]*)([a-z]*)(.+)\.qc2$ ]]
```
Fryguy, line 11: Minor, but I tend to wrap variables in `""` so that if they are blank you don't end up with a weird syntax error.  So, here, something like

```suggestion
[[ "$image_name" =~ ^(.+)-(libvirt|ovirt)-([0-9]*)([a-z]*)(.+)\.qc2$ ]]
```

scripts/smoke_test.sh
```diff
@@ -0,0 +1,258 @@
...
+product_name=${BASH_REMATCH[1]}              #manageiq
+platform=${BASH_REMATCH[2]}                  #libvirt
+product_version_number=${BASH_REMATCH[3]}    #18? (this is almost always blank, will fill in later)
+product_version_name=${BASH_REMATCH[4]}      #quinteros?
+product_version_remainder=${BASH_REMATCH[5]} #-1-20240403
+# Fill in product version number if missing
+if [[ $product_version_number == "" ]]; then
```
Fryguy, line 18: Similarly here

```suggestion
if [[ "$product_version_number" == "" ]]; then
```

scripts/smoke_test.sh
```diff
@@ -0,0 +1,258 @@
...
+  letters=({a..z})
+  letter=${product_version_name::1}
+  for (( i=0; ; i++ ))
+  do
+    if [[ ${letters[$i]} == $letter ]]; then
+      product_version_number=$(($i + 1))
+      break
+    fi
+  done
+fi
```
Fryguy, line 28: Is this basically just getting the ordinal value of the first character?  If so, I think this is a lot simpler:

```suggestion
if [[ "$product_version_number" == "" ]]; then
 product_version_number=$(($(printf '%d' "'${product_version_name::1}") - 96))
fi
```

Fryguy:
```
$ product_version_name=quinteros
$ echo $(($(printf '%d' "'${product_version_name::1}") - 96))
17
```

scripts/smoke_test.sh
```diff
@@ -0,0 +1,258 @@
...
+
+script_directory=$(dirname -- "$( readlink -f -- "$0"; )")
```
Fryguy, line 30: I don't think this works on a Mac (`readlink` has different params), but this is portable:

```bash
script_directory=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
```

scripts/smoke_test.sh
```diff
@@ -0,0 +1,258 @@
...
+
+vm_name="smoketest-${BASH_REMATCH[3]}${BASH_REMATCH[4]}${BASH_REMATCH[5]}"
```
Fryguy, line 32: should this be using the variable names instead?

```suggestion
vm_name="smoketest-$product_version_number$product_version_name$product_version_remainder"
```

scripts/smoke_test.sh
```diff
@@ -0,0 +1,258 @@
...
+
+echo "=== Waiting for SSH..."
+while true; do
+  ncat ${ip_address} 22 < /dev/null && break
+  sleep 1
+done
+sleep 1
+
+echo "=== Adding to known_hosts..."
+cat ~/.ssh/known_hosts |grep -v $ip_address > ~/.ssh/known_hosts
```
Fryguy, line 61:
```suggestion
cat ~/.ssh/known_hosts | grep -v $ip_address > ~/.ssh/known_hosts
```

scripts/smoke_test/10-evmserverd.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+echo "=== Checking evmserverd..."
+evmserverd_active=$(sshpass -p $newpassword ssh -tt root@$ip_address "systemctl is-active evmserverd")
+echo "--- Expected: active  Found: ${evmserverd_active}"
+# [[ $evmserverd_active =~ "active" ]] && exit 1 ### TODO: for some reason this is failing string and regex matching
```
Fryguy, line 6: Could it be a trailing `\r` ?  Wonder if this would work better:

```suggestion
# [[ "$evmserverd_active" =~ "active" ]] && exit 1 ### TODO: for some reason this is failing string and regex matching
```

Fryguy:
Overall, LGTM.  Great work!

scripts/smoke_test.sh
```diff
@@ -0,0 +1,258 @@
...
+### Configure the appliance
+echo "=== Configuring the appliance..."
+sshpass -p $newpassword ssh -tt root@$ip_address "hostnamectl hostname ${ip_address}.xip.io"
+sshpass -p $newpassword ssh -tt root@$ip_address "echo '${ip_address} ${ip_address}.xip.io 192' >> /etc/hosts"
+/usr/bin/expect <<EOF;
+spawn sshpass -p $newpassword ssh -tt root@$ip_address appliance_console
+match_max 100000
+expect "Press any key to continue.\r"
+send -- " "
+expect -exact "\[H\[2J\[3JAdvanced Setting\r
```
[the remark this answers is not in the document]
bdunne: No, I wasn't sure if it would be helpful in the future when a change is made to the appliance console and this needs to be edited.  I think I can reduce it and keep the important details.

scripts/smoke_test.sh
kbrock: I agree. Fun fact: the rules around `[` aka `test` and `[[` are different here:

```bash
$ [[ $x = "1" ]] && echo "yay"
$ [ $x = "1" ] && echo "yay"
-bash: [: =: unary operator expected
```

scripts/smoke_test.sh
kbrock: also, not sure if `empty string` is more readable for you:

```
$ help test
[...]
    String operators:

      -z STRING      True if string is empty.

      -n STRING
         STRING      True if string is not empty.

      STRING1 = STRING2
                     True if the strings are equal.
      STRING1 != STRING2
                     True if the strings are not equal.
```

Fryguy: Oh yeah good point.  I would use -z here possibly.

Fryguy (approved):

PR #570, opened 2024-05-01
Enable the serial console for use in kubevirt

bdunne:

### DIFF
--- a/kickstarts/partials/main/bootloader.ks.erb
+++ b/kickstarts/partials/main/bootloader.ks.erb
@@ -5,5 +5,5 @@ bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet
 <% elsif @target == "gce" %>
 bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=ttyS0,38400n8"
 <% else %>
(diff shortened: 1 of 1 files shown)


kickstarts/partials/main/bootloader.ks.erb
```diff
@@ -5,5 +5,5 @@ bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet
 <% elsif @target == "gce" %>
 bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=ttyS0,38400n8"
 <% else %>
-bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0"
+bootloader --location=mbr --driveorder=vda --append="crashkernel=auto rhgb quiet net.ifnames=0 biosdevname=0 console=ttyS0,115200 console=tty0"
```
Fryguy, line 8: Should this be conditional to kubevirt?  Or asking another way how does this affect non-kubevirt usage?

bdunne: `tty0` is the default and since it's listed last, it will continue to be the default.  Adding `ttyS0` enables the serial console as a secondary console.

PR #573, opened 2024-07-29
Stop using legacy network interface names (eth0)

bdunne:
It's no longer working for GCE and Azure appliances, they mark the connection as `unmanaged` and won't allow it to come up.  Also, our scripts no longer depend on the interface name in EL9.

net.ifnames=0 reverts to legacy adapter names (eth0)
biosdevname does the same for Dell systems and requires a package we don't have installed
network-scripts are no longer used by NetworkManager

### DIFF
--- a/kickstarts/base.ks.erb
+++ b/kickstarts/base.ks.erb
@@ -127,23 +127,8 @@ passwd -e root
 # Remove root from the cockpit disallowed-users list
 sed -i s/root//g /etc/cockpit/disallowed-users
 
-# During the image build, dracut sets up ifcfg-en<whatever> for the device.
-# On the appliances we expect eth0, so remove all the ifcfg-en<whatever> device configurations
-# and write a config for eth0 allowing the 'network' service to come up cleanly.
-rm -f /etc/sysconfig/network-scripts/ifcfg-en*
-
 # Remove machine-id
 cat /dev/null > /etc/machine-id
-
-cat <<EOF > /etc/sysconfig/network-scripts/ifcfg-eth0
-DEVICE=eth0
-BOOTPROTO=dhcp
-ONBOOT=yes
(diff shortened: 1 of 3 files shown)


Fryguy:
Thanks for figuring this out @bdunne 

cc @agrare thought you might be interested.

Fryguy (approved):

Fryguy:
LGTM, but would like a second set of eyes. @agrare ?

bdunne:
I tested removing these kernel args from both Azure and GCP appliances locally and the interfaces came up.

bdunne:
May also be related to https://github.com/ManageIQ/manageiq/issues/22973

agrare:
Love it, and great that "doing the right thing" fixes these issues also.  We do still assume eth0 in a few places that we'll need to update:

```
manageiq/app/models/miq_server/environment_management.rb:          eth0 = LinuxAdmin::NetworkInterface.new("eth0")
manageiq-appliance_console/bin/appliance_console:  eth0 = LinuxAdmin::NetworkInterface.new(NETWORK_INTERFACE)
```

agrare (approved):

agrare:
There is one more place in appliance_console with eth0, database replication:
```
lib/manageiq/appliance_console/database_replication_primary.rb:      self.primary_host      = LinuxAdmin::NetworkInterface.new(NETWORK_INTERFACE).address
lib/manageiq/appliance_console/database_replication_standby.rb:      self.standby_host      = LinuxAdmin::NetworkInterface.new(NETWORK_INTERFACE).address
lib/manageiq/appliance_console/database_replication.rb:    NETWORK_INTERFACE = 'eth0'.freeze
```

bdunne:
Ok, I think all related PRs are merged now.

PR #576, opened 2024-10-15
[QUINTEROS] Add UBI repos to pick up the newer ruby

bdunne:
CentOS Stream 8 is EOL, so packages are not getting updated anymore.  UBI8 however does still receive updates and has a newer version of ruby than CentOS Stream 8.  Adding UBI 8 repos to the appliance allows us to pull in the newer ruby so that the RPM build environment and the appliance are running the same versions of ruby.

https://github.com/msgpack/msgpack-ruby/issues/369#issuecomment-2414985034


Fryguy:
I assume you'll have another PR for upgrades?

bdunne:
Followup: https://github.com/ManageIQ/manageiq-rpm_build/pull/513

PR #587, opened 2025-08-05
Updates to deploy_kubevirt_vm

Fryguy:
@jrafanie Please review.  See individual commits for changes.

- Add simpler login check and add a storageclass check
- Add a pre-check for KubeVirt and CDI installation
- Simplify the datavolume check and report percentage complete
- Remove the correct kubevirt_known_hosts file

I could break these up into separate PRs if you want, but that's a lot of work LOL

### DIFF
--- a/bin/deploy_kubevirt_vm
+++ b/bin/deploy_kubevirt_vm
@@ -119,7 +119,7 @@ EOF
     return $?
 }
 
-create_datavolume_cr_2() {
+create_datavolume_db_cr() {
     cat << EOF | oc apply -f -
 apiVersion: cdi.kubevirt.io/v1beta1
 kind: DataVolume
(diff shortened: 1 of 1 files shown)


bin/deploy_kubevirt_vm
```diff
@@ -365,69 +365,29 @@ create_namespace_and_secret(){
...
-            if [ $(($(date +%s) - start_time)) -ge $timeout ]; then
-                log_error_timestamp "PVC binding timed out after ${timeout} seconds."
-                break
-            fi
-            sleep 5
-        done
-
-        log_timestamp_console "PVCs are not bound. Cleaning up and retrying..."
-        oc delete pvc $VM_NAME -n $NAMESPACE
-        oc delete pvc $VM_NAME-database -n $NAMESPACE
```
jrafanie, line 395: Do we need the cleanup code in case of pvc binding timeout?

jrafanie (commented):
Looking good.  I had just a minor question about the storage cleanup in case of binding timeout that was removed.

bin/deploy_kubevirt_vm
Fryguy: The old code didn't really timeout per se, it just didn't wait long enough. Instead it had an artificial timeout and then would just blow away things (even if they were in the middle of importing.

PR #598, opened 2026-03-20
Use appliance console cli for configuring single appliance or primary/standby databases

jrafanie:
* Use appliance_console_cli with ssh -c instead of expect / interactive for "configure application" (not run for standby as they only run a database for now)
* Convert set hostname expect / ssh interactive to use ssh -c
* Add replication options to the appliance_console_cli for primary/standby
* Cleanup / remove unneeded sleep / consolidate similarly named functions with one now just a wrapper around the other.

TODO:
* In a follow up, I'm going to remove the unnecessary comments and try to tidy up a bit as this PR is getting hard to follow. I did remove some of them from the commits that touched them here.
* The logging could be cleaned up so it's easier to read visually.  I started looking at making sure we have bookending of log messages here but will try to add that consistently for the rest of the code afterwards.
EDIT: (the prior two bullets are addressed in https://github.com/ManageIQ/manageiq-appliance-build/pull/599)
* There's more expect syntax being used to do interactive work that we can probably change to ssh -c.

### DIFF
--- a/bin/deploy_kubevirt_vm
+++ b/bin/deploy_kubevirt_vm
@@ -959,25 +959,6 @@ setup_vm_resources() {
 }
 
 
-# Function to configure MIQ appliance
-# Parameters:
-#   $1: replication_type - "none", "primary", or "standby"
-#   $2: cluster_node_number - node number (1 for primary, 2 for standby)
-#   $3: primary_host_ip - IP address of primary host (required for standby)
-#   $4: standby_host_ip - IP address of standby host (required for standby)
-appliance_configure() {
-    local replication_type=${1:-"none"}
-    local cluster_node_number=${2:-""}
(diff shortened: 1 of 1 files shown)


bin/deploy_kubevirt_vm
```diff
@@ -1001,25 +978,43 @@ fi
...
-appliance_configure
-
-log "First VM configuration completed successfully."
+# Store primary VM info for HA configuration
 primary_vmi_name=$vmi_name
+primary_vmi_ip=$vmi_ip
 
 # Deploy second VM if HA mode
 if [[ "$HA" == "true" ]]; then
+    # Configure primary
```
jrafanie: I'm removing some of these useless comments in a followup as they're also in areas not touched by this PR.

Fryguy (approved):

